PR #85, opened 2021-10-11
Fix: issue with user:login API, returning password in the response

Nitinrajyadav:


chenos:
fixed https://github.com/nocobase/nocobase/commit/d94606ab302f0a00877211e2da84934f611f7948

@Nitinrajyadav  Thanks for the feedback. The getter method doesn't work. Another way is to add a default scope to the user collection.

```ts
{
  defaultScope: {
    attributes: { exclude: ['password'] },
  }
}
```

PR #114, opened 2021-11-29
feat: allow sign up

semmywong:
feat: allow sign up


packages/plugin-system-settings/src/collections/system_settings.ts
```diff
@@ -11,6 +11,10 @@ export default {
       type: 'boolean',
       name: 'showLogoOnly',
     },
+    {
+      type: 'boolean',
+      name: 'allowSignUp',
```
chenos, line 16: 少了个默认值 defaultValue: true

PR #125, opened 2021-12-04
refactor: table cell text overflow that show ellipsis

semmywong:
Refactor/table: table cell text overflow that show ellipsis


packages/client/src/schemas/table/hooks/useTableColumns.tsx
```diff
@@ -22,6 +22,7 @@ export const useTableColumns = () => {
 
   const columnSchemas = schema.reduceProperties((columns, current) => {
     if (isColumn(current)) {
+      current['x-component-props']['ellipsis'] = true;
```
chenos, line 25: 这里需要判断一下 x-component-props 不存在的情况

semmywong: 已修复

PR #198, opened 2022-02-18
fix: postgres sort with appends issue

chareice:


packages/plugin-collection-manager/src/__tests__/action.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  let db: Database;
+  let app: MockServer;
+
+  beforeEach(async () => {
+    app = mockServer();
+    app.plugin(CollectionManagerPlugin);
+    app.plugin(UiSchemaStoragePlugin);
+
+    db = app.db;
+    await db.clean({ drop: true });
```
chenos, line 16: db.clean 去掉

Issue #809, opened 2022-09-05
Rounding in the Number field to 6 significant digits


zhouyanliang:
### Discussed in https://github.com/nocobase/nocobase/discussions/801

<div type='discussions-op-text'>

<sup>Originally posted by **arzanov** September  2, 2022</sup>
When I enter a number longer than 6 digits, I get rounding in the Number field to 6 significant digits. The precision is set to 1.00, but if the integer part is longer than 4 digits, rounding after the decimal point occurs. If I enter an integer longer than 6 digits in this field, then it is rounded up to 6 significant digits, and the rest becomes 0.
Why is this happening and how to fix it?</div>

PR #810, opened 2022-09-06
fix: number storage type changed to double

chenos:


chenos:
discussion: https://github.com/nocobase/nocobase/discussions/801
issue: https://github.com/nocobase/nocobase/issues/809

PR #827, opened 2022-09-14
Russian translation

arzanov:


arzanov:
I found some bugs in the code

PR #906, opened 2022-10-13
feat: add excel function field

azriel46d:
This PR follows similar implementation to the `Formula` field 

1. It introduces https://www.npmjs.com/package/hot-formula-parser . This package was chosen over `@formulajs/formulajs` due to the string parsing (using jison) rather than an `eval` function. I considered as well https://hyperformula.handsontable.com/ however it seemed more complex for the purpose at hand, unless this is added to table views. 
2. Similar to the `formula` field it keeps the `@` to load the fields. 
3. When evaluating the scope is passed in as variables to `hot-formula-parser`

image

https://user-images.githubusercontent.com/2339134/195577579-2e7fd202-01e1-41c1-8b90-d6bf6f1a9646.mov


azriel46d:
Also one more suggestion to help the users, would be to introduce better intellisense, would it make sense to introduce something like the [monaco-editor](https://www.npmjs.com/package/monaco-editor) ?

chenos:
> I've added a helper function to consolidate the genration of a parser , setting of variables and extra functions. however, path wise i'm sure its not in the best place to be done. Could you kindly indicate where such a helper should be placed when its shared between the db folder and the schema folder?

Can be put in `@nocobase/utils` like this https://github.com/nocobase/nocobase/commit/6aa44ec539a3f930274a29220d7d13ac563c9593

Server

```ts
import { getFormulaParser } from '@nocobase/utils';
```

Client

```ts
import { getFormulaParser } from '@nocobase/utils/client';
```

chenos:
> Also one more suggestion to help the users, would be to introduce better intellisense, would it make sense to introduce something like the [monaco-editor](https://www.npmjs.com/package/monaco-editor) ?

monaco-editor is overqualified and bulky. Any other better suggestions?

chenos:
The expected results are `2+1.22=3.22`

image

azriel46d:
> The expected results are `2+1.22=3.22`
> 
> image
Browser console does exactly the same . I believe it has to do with Javascript precision. To which case the suggestion would be to use the mathjs and formula option which possibly as a library handles these better
image

chenos:
```ts
2 + 1.11
// 3.1100000000000003

import math from 'mathjs';
math.evaluate('2+1.11')
// 3.1100000000000003
math.round(2 + 1.11, 9)
// 3.11

import { Parser } from 'hot-formula-parser';
import math from 'mathjs';

const parser = new Parser();
const r = parser.parse('SUM(2, 1.11)'); 
// 3.1100000000000003
math.round(r.result, 9)
// 3.11
```

It can be solved with `math.round(result, 9)`

azriel46d:
> ```ts
> 2 + 1.11
> // 3.1100000000000003
> 
> import math from 'mathjs';
> math.evaluate('2+1.11')
> // 3.1100000000000003
> math.round(2 + 1.11, 9)
> // 3.11
> 
> import { Parser } from 'hot-formula-parser';
> import math from 'mathjs';
> 
> const parser = new Parser();
> const r = parser.parse('SUM(2, 1.11)'); 
> // 3.1100000000000003
> math.round(r.result, 9)
> // 3.11
> ```
> 
> It can be solved with `math.round(result, 9)`

True, but not all functions in Excel have a numeric result. 
Take the case of having `IF(OR(x = 'hello', x = 'world'), true, false)` 
If Rounding is needed I would suggest `ROUND(SUM(2,1.11), 1)` which would have the same result. In this manner the user has full flexibility on how to use the result.

PR #1234, opened 2022-12-12
chore: upgrade sequelize version to to latest

Dunqing:


muthomihian:
You should come up with a language translator AI

PR #1272, opened 2022-12-20
perf: no need to query sort field every time

semmywong:
no need to query sort field every time


chenos:
已改进 https://github.com/nocobase/nocobase/commit/83bca97db0030b476d61c34c59b6c56f179e29b9

PR #1344, opened 2023-01-10
feat: formula plugin

chenos:


chenos:
Thanks to @arzanov, the excel formula plugin was merged.

PR #1514, opened 2023-02-28
feat: association filter improve

anuoua:
- add data scope for association filter
- add default sort for association filter


packages/core/client/src/schema-component/antd/association-filter/AssociationFilter.Initializer.tsx
```diff
@@ -21,7 +21,10 @@ export const AssociationFilterInitializer = () => {
       name: field.name,
       title: field.uiSchema.title,
       type: 'void',
+      'x-target': field.target,
       'x-designer': 'AssociationFilter.Item.Designer',
+      'x-decorator': 'AssociationFilter.Item.Decorator',
```
chenos, line 26: 没考虑兼容性问题，这里改成这样，旧的 schema 就挂掉了

PR #1744, opened 2023-04-21
feat: deleted collection field give some tips

lainlee:
When delete field in collections, show a tip **The field has bee deleted** in the past block container


zhangzhonghe:
@hongboji 这个是否需要测试下？

PR #2034, opened 2023-06-12
chore(database): return emtpy fields when attributes not specified

chareice:


chareice:
Link: T-428

chareice:
Link T-428

PR #2066, opened 2023-06-16
fix: association field cannot enable link in table column

katherinehhh:
## Description (Bug 描述)
关系字段在表格中无法配置启用链接


katherinehhh:
Link T-529

PR #2070, opened 2023-06-18
refactor(db): add batch logic to update for better performance

mytharcher:
# Description (需求描述)

add batch logic to `repository.update` for better performance.

# Motivation (需求背景)

Before this PR, `repository.update` only update rows one by one (DB IO), mostly for triggering single model events. But when used on large quantity of data, it will be very slow.

So, we need better performance in such scenario, which only send one IO to DB for multiple records updating, and not triggering single model events.

# Key changes (关键改动）

- Frontend (前端)
  - Add `individualHooks` in update node configuration in workflow.
  - Add linkage for values in update node configuration to skip non-belongsTo association values.
- Backend (后端)
  - Add branch logic on `individualHooks` is exact `false` to perform batch updating.
  - Fix result value of update node to only numbers.

# Test plan (测试计划)

## Suggestions (测试建议)

* Unit cases.
* Benchmark.

## Underlying risk (潜在风险)

Not support `hasOne`, `hasMany` and `belongsToMany` values in batch updating.

# Showcase (结果展示）

53732540-fec5-473b-abb8-fcb599405582

Only spent about 400ms for updating 20000 records in workflow, instead of minutes which reported by client.

### DIFF
--- a/packages/core/database/src/repository.ts
+++ b/packages/core/database/src/repository.ts
@@ -609,6 +609,45 @@ export class Repository<TModelAttributes extends {} = any, TCreationAttributes e
 
     const queryOptions = this.buildQueryOptions(options);
 
+    // NOTE:
+    // 1. better to be moved to separated API like bulkUpdate/updateMany
+    // 2. strictly `false` comparing for compatibility of legacy api invoking
+    if (options.individualHooks === false) {
+      const { model: Model } = this.collection;
+      // @ts-ignore
+      const primaryKeyField = Model.primaryKeyField || Model.primaryKeyAttribute;
+      // NOTE:
+      // 1. find ids first for reusing `queryOptions` logic
+      // 2. estimation memory usage will be N * M bytes (N = rows, M = model object memory)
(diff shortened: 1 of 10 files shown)


mytharcher:
Link T-541.

packages/core/database/src/repository.ts
```diff
@@ -609,6 +609,45 @@ export class Repository<TModelAttributes extends {} = any, TCreationAttributes e
...
+          );
+        }),
+        transaction,
+      });
+      const [result] = await Model.update(values, {
+        where: {
+          [primaryKeyField]: rows.map((row) => row.get(primaryKeyField)),
+        },
+        fields: options.fields,
+        hooks: options.hooks,
```
chareice, line 640: 这个hooks参数是否需要默认设置为false

mytharcher: 如果默认是 `false`，以后需要 `afterBulkUpdate` 这类事件也就没有触发了，这里不应该限制，API 调用者自行决定比较好。

PR #2089, opened 2023-06-21
fix: unable to load data from chinaRegion during the first configuration

katherinehhh:
## Description (Bug 描述)
中国行政区数据第一次配置的时候加载不出来，刷新后正常


chenos:
Close T-607

PR #2632, opened 2023-09-12
refactor(plugin-workflow): add ready state for gracefully tearing down

mytharcher:
### Steps to reproduce (复现步骤)

1. Trigger a workflow and immediately stop server.
2. Start the server and immediately reload.

### Expected behavior (预期行为)

Triggered workflow should run when server is ready. If stopped or reloaded, should gracefully tearing down.

### Actual behavior (实际行为)

Still running workflow after server stopped and throw error about database.

## Related issues (相关 issue)

None.

## Reason (原因)

No cleaning in tear down period.

## Solution (解决方案)

Do cleaning in tear down period.


mytharcher:
Merged in #2430.

PR #2948, opened 2023-11-01
New Plugin-> NumberToWords

codenuladev:
# Description (需求描述)
Add a new plugin called plugin-number-to-words that allows users to compute values from other fields and convert those numeric values to words. It can also convert numbers to currency and apply decimal checks, making it ideal for a wide range of applications.

# Motivation (需求背景)
This plugin would be useful for a variety of tasks, such as generating invoices and receipts, creating business proposals and reports, writing financial reports and analyses, and creating technical documentation.

# Key changes (关键改动）
Provide a technically detailed description of the key changes made.
- Frontend (前端)-
1)Added a new field 'numberToWords' in the Add new fields dropdown,
2)Added a dropdown in the drawer with currency values (ex-India-hindi,USA-English etc)
3)Added decimal check checkbox to take account of decimals while converting numeric values to words.
4)Added a 'Do not add only' checkbox to hide/show only after conversion of the numeric values in case of currency selection
5)Added an expression box to operate on the selected fields from the variable (ex sum1+sum2)
6)Expression.tsx includes Variable.TextArea to select variables/Fields to operate upon (ex sum1+sum2).
7)Result.tsx contains the logic of computing results according to the expression generated by Expression.tsx.

- Backend (后端)
1)In numberToWordsFields.ts, we defines a custom field type called NumberToWordsField. 
This field type allows users to store numbers in words in their database. 
It works by using the evaluators library to evaluate a user-provided expression to calculate 
the value of the field. The expression can be used to perform any mathematical operation, including 
converting a number to words.

2)The NumberToWordsField class has a number of options that can be used to customize its behavior. 
These options include:
engine: The engine to use to evaluate the expression. The default engine is math.js.
expression: The expression to evaluate to calculate the value of the field.
dataType: The data type to store the value of the field in. The default data type is string.
currency: The currency to use when formatting the value of the field.
uiSchema: A schema that can be used to customize the appearance and behavior of the field in the NocoBase user interface.

3)The NumberToWordsField class also has a number of methods that can be used to manage and interact with the field. 
These methods include:
calculate(): Calculates the value of the field based on the expression and scope provided.
initFieldData(): Initializes the field data for all records in the collection.
calculateField(): Calculates the value of the field for the specified instance.
updateFieldData(): Updates the field data for all records in the collection.
# Test plan (测试计划)
## Suggestions (测试建议)
Provide any suggestions or recommendations for improvements in the testing plan.

## Underlying risk (潜在风险)
Identify any potential risks or issues that may arise from the new feature or modification.

# Showcase (结果展示）
Including any screenshots of the new feature or modification.


chenos:
You can create a repository to maintain this plugin by yourself. Thank you!

PR #2982, opened 2023-11-07
fix(formula): fix result scope

mytharcher:
## Description (Bug 描述)

Formula field in form not react when referenced fields changes.

### Steps to reproduce (复现步骤)

1. Add a table with formula field.
2. Add another field to be used in formula.
3. Add a form to add new record, and add the 2 fields.
4. Change value by inputing for the referenced field.

### Expected behavior (预期行为)

The result of formula field should change after.

### Actual behavior (实际行为)

Not changes.

## Related issues (相关 issue)

#2449.

## Reason (原因)

Didn't replace in root form:

```
const fieldPath = path?.replace(`.${fieldSchema.name}`, '');
```

## Solution (解决方案)

Change the replace logic.


mytharcher:
Use #2983 instead.

PR #3106, opened 2023-11-28
fix(plugin-workflow): fix form event parse variables

mytharcher:
## Description (Bug 描述)

Cannot get variable assigned value from parent record in form workflow.

### Steps to reproduce (复现步骤)

1. Add a form workflow.
2. Bind the workflow form an associating collection sub-table creation form on "Submit to workflow" button.
3. Add assigned value with variable pointing to parent record of associating collection.
4. Submit the form and trigger workflow.

### Expected behavior (预期行为)

Could get parent record value in context.

### Actual behavior (实际行为)

Could not.

## Related issues (相关 issue)

None.

## Reason (原因)

Variable values not parsed before submit.

## Solution (解决方案)

Use abstracted value collecting method to do same parsing as creating record.

### DIFF
--- a/packages/core/client/src/block-provider/hooks/index.ts
+++ b/packages/core/client/src/block-provider/hooks/index.ts
@@ -4,7 +4,7 @@ import { App, message } from 'antd';
 import _, { cloneDeep } from 'lodash';
 import get from 'lodash/get';
 import omit from 'lodash/omit';
-import { ChangeEvent, useContext, useEffect } from 'react';
+import { ChangeEvent, useCallback, useContext, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


packages/core/client/src/block-provider/hooks/index.ts
```diff
@@ -97,85 +97,109 @@ export function getFormValues({
...
+    });
+
+    const assignedValues = {};
+    const waitList = Object.keys(originalAssignedValues).map(async (key) => {
+      const value = originalAssignedValues[key];
+      const collectionField = getField(key);
+
+      if (process.env.NODE_ENV !== 'production') {
+        if (!collectionField) {
+          throw new Error(`useCreateActionProps: field "${key}" not found in collection "${name}"`);
```
zhangzhonghe, line 132: 报错信息建议改成：field "${key}" not found in collection "${name}"

PR #3435, opened 2024-01-25
feat(core): add message handling on success

mytharcher:
# Description (需求描述)

Add message handling on success.

# Motivation (需求背景)

If need to show some messages from successful response, there is no way.

# Key changes (关键改动）

- Frontend (前端)
  - Response interceptor in APIClient.
- Backend (后端)
  - Data wrapping middleware.

# Test plan (测试计划)

## Suggestions (测试建议)

None.

## Underlying risk (潜在风险)

None.

# Showcase (结果展示）

None.

### DIFF
--- a/packages/core/server/src/middlewares/data-wrapping.ts
+++ b/packages/core/server/src/middlewares/data-wrapping.ts
@@ -56,6 +56,10 @@ export function dataWrapping() {
         data: ctx.body,
       };
     }
+
+    if (ctx.state.messages?.length) {
+      ctx.body.messages = ctx.state.messages;
+    }
   };
 }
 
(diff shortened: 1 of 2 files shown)


packages/core/server/src/middlewares/data-wrapping.ts
```diff
@@ -56,6 +56,10 @@ export function dataWrapping() {
         data: ctx.body,
       };
     }
+
+    if (ctx.state.messages?.length) {
+      ctx.body.messages = ctx.state.messages;
```
chenos, line 61: 如果 body 是 null 这里就报错了？

PR #4103, opened 2024-04-19
fix(linkage-rule): linkage rule support empty condiction

katherinehhh:
# Description

# Motivation

# Key changes

- Frontend
- Backend

# Test plan
## Underlying risk

# Showcase


katherinehhh:
LInk T-4091

PR #4201, opened 2024-04-27
fix(plugin-workflow): fix duplicated triggering schedule event in multiple apps

mytharcher:
## Description

Fix duplicated triggering schedule event in multiple apps.

### Steps to reproduce

1. Enable multiple app plugin and multi-app share collection plugin.
2. Create a sub-app.
3. Create a schedule event starts on static time and enable it.

### Expected behavior

The workflow only be triggered once each time (if repeated).

### Actual behavior

Triggered as many times as how many apps.

## Related issues

None.

## Reason

Duplicated event in different app instance.

## Solution

Add unique key (`eventKey`) to executions table, and generate it in triggers.

### DIFF
--- a/packages/plugins/@nocobase/plugin-workflow/src/server/Plugin.ts
+++ b/packages/plugins/@nocobase/plugin-workflow/src/server/Plugin.ts
@@ -1,8 +1,9 @@
 import path from 'path';
+import { randomUUID } from 'crypto';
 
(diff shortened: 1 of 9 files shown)


baiheinet:
终于解决了*★,°*:.☆(￣▽￣)/$:*.°★* 。

PR #4468, opened 2024-05-23
refactor: export & import plugin

chareice:


packages/core/client/src/appInfo/CurrentAppInfoProvider.tsx
```diff
@@ -22,6 +22,7 @@ export const useCurrentAppInfo = () => {
       };
       lang: string;
       version: string;
+      exportLimit?: number;
```
amondirvin, line 25: 1000000.00

amondirvin (approved):

packages/core/client/src/appInfo/CurrentAppInfoProvider.tsx
chareice: This will cause the entire application to hang and not be able to handle other requests

PR #4506, opened 2024-05-28
feat(client): add new variable named 'URL search params' and support link action

zhangzhonghe:
close T-4367
close T-4368

Unit tests：
- [x] Override the code in the Link button that converts the value entered by the user into a query string.
- [x] Override the code in the variable that converts the URL query string into the variable context

E2E:
- [x] Measure a basic and complete process, which consists of: 1. full parameter input on the Link side (jumping to the current page); 2. data scope is normal; 3. data is re-requested after clicking a different Link button.

# Add new variable

Add a new variable called "URL search params".

image

## How to use

This variable will only work if there is a query string in the URL of the page, otherwise it will be disabled.

image

# Filter form fields support setting default values

image

# Add new action named 'Link'

This action supports configuring the URL of the page to jump to.

e3b3f361-6e7c-4f0c-b9b0-89ca8e5155dc


zhangzhonghe:
Changes to #4508 were merged here.

PR #4603, opened 2024-06-08
feat(tree-block): support filtering child nodes

2013xile:
# Description

Support filtering child nodes in tree collection block.

# Motivation

fix T-4497

# Key changes

- Frontend
- Backend
  - Set `tree=true` to `tree=false` as `filter` parameter is not empty when filtering tree collection.

# Test plan
## Underlying risk

# Showcase

https://github.com/nocobase/nocobase/assets/3250534/bd34d25e-c7c7-47ef-8f55-feb5e3bd8e40

### DIFF
--- a/packages/core/data-source-manager/src/default-actions/list.ts
+++ b/packages/core/data-source-manager/src/default-actions/list.ts
@@ -10,6 +10,7 @@
 import { assign } from '@nocobase/utils';
 import { pageArgsToLimitArgs } from './utils';
 import { Context } from '@nocobase/actions';
+import _ from 'lodash';
 
 function totalPage(total, pageSize): number {
   return Math.ceil(total / pageSize);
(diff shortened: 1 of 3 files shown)


packages/core/data-source-manager/src/default-actions/list.ts
```diff
@@ -20,11 +21,15 @@ function findArgs(ctx: Context) {
   const params = ctx.action.params;
 
   if (params.tree) {
-    const [collectionName, associationName] = resourceName.split('.');
-    const collection = ctx.db.getCollection(resourceName);
-    if (collection.options.tree && !(associationName && collectionName === collection.name)) {
-      const foreignKey = collection.treeParentField?.foreignKey || 'parentId';
-      assign(params, { filter: { [foreignKey]: null } }, { filter: 'andMerge' });
+    if (!_.isEmpty(params.filter)) {
```
chenos, line 24: 判断 filter 用这个
https://github.com/nocobase/nocobase/blob/T-4497/packages/core/utils/src/isValidFilter.ts

PR #4609, opened 2024-06-09
 feat(client): add searchbar and increase ux of select popover in icon-picker component

leptoquark1:
# Description
- Adds the ability for the user to search for icons
- Increase the UX by increasing the icon sizes and add a title tag to display the icon name on mousehover

# Motivation
- No way to search for a specific icon in icon-select popover
- To small icons make it hard to spot the correct item

# Key changes
- Frontend
  - Add input of type search to icon-select popover
  - Filter the available icons by comparing search input with icon name
  - Debounce the icon search filter to maintain performance on weak client devices
  - Add title tag to icon component wrapper element

# Showcase

image

### DIFF
--- a/packages/core/client/src/schema-component/antd/icon-picker/IconPicker.tsx
+++ b/packages/core/client/src/schema-component/antd/icon-picker/IconPicker.tsx
@@ -35,35 +35,21 @@ interface IconPickerReadPrettyProps {
 }
 
 const useStyle = (isSearchable: IconPickerProps['searchable']) =>
-  createStyles(({ token, css }) => {
+  createStyles(({ css }) => {
     return {
       popoverContent: css`
         width: 26em;
         ${!isSearchable && 'max-'}height: 20em;
         overflow-y: auto;
       `,
-      searchbarDivider: css`
-        margin-top: ${token.marginXS}px;
-        margin-bottom: ${token.margin}px;
-
-        &::after {
-          position: relative;
-          top: 12px;
-          display: block;
-          width: calc(100% - 20px);
-          height: 1px;
-          background: ${token.colorSplit};
-          content: '';
-        }
-      `,
     };
   })();
 
 function IconField(props: IconPickerProps) {
-  const { fontSizeXL } = theme.useToken().token;
+  const { fontSizeHeading3 } = theme.useToken().token;
   const availableIcons = [...icons.keys()];
   const layout = useFormLayout();
-  const { value, onChange, disabled, iconSize = fontSizeXL, searchable = true } = props;
+  const { value, onChange, disabled, iconSize = fontSizeHeading3, searchable = true } = props;
   const [visible, setVisible] = useState(false);
   const [filteredIcons, setFilteredIcons] = useState(availableIcons);
   const { t } = useTranslation();
@@ -115,16 +101,15 @@ function IconField(props: IconPickerProps) {
             <div>
               <div>{t('Icon')}</div>
               {searchable && (
-                <div className={styles.searchbarDivider}>
-                  <Search
-                    role="search"
-                    name="icon-search"
-                    placeholder={t('Search') + '...'}
-                    allowClear
-                    onSearch={filterIcons}
-                    onChange={(event) => filterIcons(event.target?.value)}
-                  />
-                </div>
+                <Search
+                  style={{ marginTop: 8 }}
+                  role="search"
+                  name="icon-search"
+                  placeholder={t('Search')}
+                  allowClear
+                  onSearch={filterIcons}
+                  onChange={(event) => filterIcons(event.target?.value)}
+                />
               )}
             </div>
           }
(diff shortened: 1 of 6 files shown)


2013xile:
Good idea. It would be better to add `searchable` and `fontSize` as properties of the `IconPicker` component, allowing for customization based on specific use cases. Additionally, using the `Input` component from `antd` instead of the original one  would maintain UI consistency.

leptoquark1:
> Good idea. It would be better to add `searchable` and `fontSize` as properties of the `IconPicker` component, allowing for customization based on specific use cases. Additionally, using the `Input` component from `antd` instead of the original one would maintain UI consistency.

I used `searchable = true` and `iconSize = 24` as default values so that the orginal intention of this PR is preserved.

packages/core/client/src/schema-component/antd/icon-picker/IconPicker.tsx
```diff
@@ -11,28 +11,45 @@ import { CloseOutlined, LoadingOutlined } from '@ant-design/icons';
...
 
 interface IconPickerReadPrettyProps {
   value?: string;
 }
 
 function IconField(props: IconPickerProps) {
+  const availableIcons = [...icons.keys()];
   const layout = useFormLayout();
-  const { value, onChange, disabled } = props;
+  const { value, onChange, disabled, iconSize = 24, searchable = true } = props;
```
2013xile, line 39: For default `iconSize`, it might be better to use a font size token like `fontSizeXL` instead of a fixed number so that it could be more comfortable when user changing the font size of the full site via theme editor.

```
import { theme } from 'antd';

// ...
const { token } = theme.useToken();
token.fontSizeXL
```

Reference: https://ant.design/docs/react/customize-theme/#design-token

packages/core/client/src/schema-component/antd/icon-picker/__tests__/icon-picker.test.tsx
```diff
@@ -47,4 +47,15 @@ describe('IconPicker', () => {
     await userEvent.click(icons[1]);
     expect(screen.queryAllByRole('img').length).toBe(0);
   }, 300000);
+
+  it('should filter the displayed icons when changing the value of search input', async () => {
+    const { container } = render(<App />);
+
+    const button = container.querySelector('button') as HTMLButtonElement;
+    await userEvent.click(button);
+
```
2013xile, line 56: image

Try to add `await waitFor()` here, waiting util the IconPicker component rendered.

packages/core/client/src/schema-component/antd/icon-picker/__tests__/icon-picker.test.tsx
```diff
@@ -47,4 +47,15 @@ describe('IconPicker', () => {
...
+    const searchInput = container.querySelector('input[name="icon-search"]') as HTMLInputElement;
+    fireEvent.change(searchInput, { target: { value: 'left' } });
```
2013xile, line 58: Is it better to use `await userEvent.type()` instead of `fireEvent.change`?

2013xile (changes requested):
1. It would be better to display a empty box when the search result is empty. https://ant.design/components/empty
2. Now the search bar is out when scrolling down, is it better to separate it with the container of icons? like this:
image


PS:
As we made a minor revision of the CLA, it needs to be re-signed. Thank you!

packages/core/client/src/schema-component/antd/icon-picker/IconPicker.tsx
leptoquark1: In this case I'd prefer `fontSizeHeading3`, since even the xlarge fontsize is too small when dealing with a bunch of icons side-by-side.

packages/core/client/src/schema-component/antd/icon-picker/__tests__/icon-picker.test.tsx
leptoquark1: I just tried to adapt the steps from existing code, and not see the actual result, because I've trouble running vitest atm on this windows machine.

leptoquark1:
@2013xile Popd some changes as suggested here

packages/core/client/src/schema-component/antd/icon-picker/__tests__/icon-picker.test.tsx
leptoquark1: See [this](https://github.com/nocobase/nocobase/pull/4609#discussion_r1641523670)

packages/core/client/src/schema-component/antd/icon-picker/IconPicker.tsx
2013xile: It is fine to try `fontSIzeHeading3` to compare the effect, I just gave an example.
And I think the divider could be removed? If it is required, there is also a component you can use. https://ant.design/components/divider

packages/core/client/src/schema-component/antd/icon-picker/__tests__/icon-picker.test.tsx
2013xile: How did you run vitest and do you have the error messages or screenshots? I may help to fix it.

leptoquark1: @2013xile I've got them running by stoping serving with `yarn dev` first, thanks.

However, I'm having problems fixing the tests because I can't seem to get the searchbar in the popover selected. Hope you can help me here

PR #6127, opened 2025-01-23
refactor(plugin-file-manager): move destroy to collection event

mytharcher:
### This is a ...

- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Others

### Motivation

Allow to delete files when file (attachment) record is deleted.

### Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Allow to delete files when file (attachment) record is deleted |
| 🇨🇳 Chinese | 支持删除文件记录时同时删除文件 |

### Docs

| Language   | Link |
| ---------- | --------- |
| 🇺🇸 English |      |
| 🇨🇳 Chinese |    |

### Checklists


thknx:
Hi there,

I tried to upload a new file and delete the record, but the situation is same. File remans in storage even if I delete it's record.

image

mytharcher:
Hi @thknx ,

What are your steps?

BTW, this feature will only work for file collection. Because there is no way to delete record of attachments collection for now (will be changed in future).

thknx:
1. I designed a Form which ask name and an attachment.
2. Insert a new record (Name = abc and a file)
3. Then press the delete button.

Record is deleted but the file remains in storage as you see above screen shot.

mytharcher:
> 1. I designed a Form which ask name and an attachment.
> 2. Insert a new record (Name = abc and a file)
> 3. Then press the delete button.
> 
> Record is deleted but the file remains in storage as you see above screen shot.

According to your case, delete the record of your custom collection will not delete the record of attachments collection. So the file will not be deleted.

The key point of deleting file is record deletion of file (attachment) collection. Not the record deletion of associating collection, which plays as referencing.

So, the only workaround for now is to use file collection to store your files (as attachment). And then use association field to the file collection in your custom collection. When you want to delete some file, do deletion from the file collection.

PR #6434, opened 2025-03-12
Updating italian translations

N3tN00b3r:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [ ] Bug fix
- [X] Translation

### Description 
I am reworking all the Italian translations of the plugins, as in many cases the original en-US file has missing terms compared to the zh-CN file. The "workflow" section is still missing and requires more effort.

### Disclaimer
I don't speak Chinese and could only do a MT based on context. I apologize in advance if there are any inaccuracies.

### Checklists
- [X] All changes have been self-tested and work as expected
- [X] Test cases are updated/provided or not needed
- [X] Doc is updated/provided or not needed
- [X] Component demo is updated/provided or not needed
- [X] Changelog is provided or not needed
- [X] Request a code review if it is necessary


Albert-mah:
Really appreciate your effort and hard work on this. Thanks so much~ 
 We'll also be speeding up making the project more localization-friendly overall, and thanks again for your awesome contribution!

PR #6435, opened 2025-03-12
feat: add German localization files for NocoBase core and plugins

jnsflschr:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [ ] Bug fix
- [x] Translation

### Motivation

There were no German translation and the buttons and messages can't be translated via the localisations plugin.

### Description 

Created German (de-DE) translation files for nocobase core and all nocobase plugins. I used the English translation and Chinese translation as reference. The German translation files contain the union of the keys from both translation files (en-EN and zh-CN) because I didn't now which were the more recent and complete translation. Sadly I couldn't test the changes because I didn't get  the local Dockerfile to work because of unrelated issues with yarn in the container. I would recommend to check the button sizes before merging because of slightly longer German words for some translation than in English or Chinese (ex. "Edit" -> "Bearbeiten").

### Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  added German translation files  |
| 🇨🇳 Chinese |           |

### Docs

| Language   | Link |
| ---------- | --------- |
| 🇺🇸 English |      |
| 🇨🇳 Chinese |    |

### Checklists
- [ ] All changes have been self-tested and work as expected
- [x] Doc is not needed


Albert-mah:
Thanks a ton for contributing German localization!🎉

PR #6588, opened 2025-03-29
fix(database): fix datetimeNoTz field

mytharcher:
### This is a ...

- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Motivation

Avoid "datetimeNoTz" field changes when value not changed in updating record.

### Description 

Sequelize is using `_.isEqual` to compare old value and new value. While the implementation in legacy `datetimeNoTz` field always set string value to record, this will cause changed (to Date value), which is not expected.

### Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Avoid "datetimeNoTz" field changes when value not changed in updating record |
| 🇨🇳 Chinese | 避免“日期时间（无时区）”字段在值未变动的更新时触发值改变 |

### Docs

| Language   | Link |
| ---------- | --------- |
| 🇺🇸 English |      |
| 🇨🇳 Chinese |    |

### Checklists

### DIFF
--- a/packages/core/database/src/fields/datetime-no-tz-field.ts
+++ b/packages/core/database/src/fields/datetime-no-tz-field.ts
@@ -78,9 +78,7 @@ export class DatetimeNoTzField extends Field {
         const momentVal = moment(val);
         if ((typeof val === 'string' && isIso8601(val)) || val instanceof Date) {
           momentVal.utcOffset(timezone);
-          if (isPg) {
-            momentVal.utcOffset(-dateOffset, true);
-          }
+          momentVal.utcOffset(-dateOffset, true);
         }
 
         if (isMySQLCompatibleDialect) {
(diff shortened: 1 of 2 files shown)


packages/core/database/src/fields/datetime-no-tz-field.ts
```diff
@@ -57,36 +55,39 @@ export class DatetimeNoTzField extends Field {
...
 
-        if (val && val instanceof Date) {
-          // format to YYYY-MM-DD HH:mm:ss
-          const momentVal = moment(val).utcOffset(timezone);
-          val = momentVal.format('YYYY-MM-DD HH:mm:ss');
+        const dateOffset = new Date().getTimezoneOffset();
+        const momentVal = moment(val);
+        if ((typeof val === 'string' && isIso8601(val)) || val instanceof Date) {
+          momentVal.utcOffset(timezone);
+          momentVal.utcOffset(-dateOffset, true);
```
chenos, line 81: 加这个是什么情况

mytharcher: 需要再偏移当前所在时区与 UTC 的时区差值，在 toDate 的时候才能得到与无时区生成的 Date 相等。不加这个原来的 test case 第二个跑不过。

PR #6704, opened 2025-04-18
fix: integer overflow when attachment size is larger than 2 GB

erkan-vatan:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Description 

Previous PR #6436 changed maximum allowed attachment size from 1 GB to infinity. But attachment size in bytes is stored in the "attachments" table with integer data type. This causes PostgreSQL to throw errors when attachment size is larger than 2 GB.
### Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix integer overflow issue when attachment size is larger than 2 GB |
| 🇨🇳 Chinese |  |

### Docs

| Language   | Link |
| ---------- | --------- |
| 🇺🇸 English |      |
| 🇨🇳 Chinese |    |

### Checklists
- [ ] Changelog is provided or not needed

### DIFF
--- a/packages/plugins/@nocobase/plugin-file-manager/src/server/collections/attachments.ts
+++ b/packages/plugins/@nocobase/plugin-file-manager/src/server/collections/attachments.ts
@@ -38,7 +38,7 @@ export default defineCollection({
     },
     {
       comment: '文件体积（字节）',
-      type: 'integer',
+      type: 'bigInt',
       name: 'size',
     },
     // TODO: 使用暂不明确，以后再考虑


mytharcher (changes requested):
Hi @erkan-vatan , thanks for your contribution!

Changing field type needs to do migration on all existing file collections. Such as `attachments` (built-in) and other file collections created from file collection template.

Here is an example:

https://github.com/nocobase/nocobase/blob/main/packages/plugins/%40nocobase/plugin-file-manager/src/server/migrations/20250225225034-increase-url-length.ts

And test cases are also needed for the migration. For example:

https://github.com/nocobase/nocobase/blob/main/packages/plugins/%40nocobase/plugin-file-manager/src/server/__tests__/migrations/20250225225034-increase-url-length.test.ts

Issue #8549, opened 2026-02-01
REST API: Details block calling List action instead of Get action


mrkmcnamee:
## * Describe the bug

The Details block calls the REST API **List** action instead of the **Get** action. This results in a legitimate request to the collection List endpoint, but the response will not contain all of the information that the Get action endpoint provides.

## * Environment

- NocoBase version: v2.0.0-beta.17

- Database type and version: PostgreSQL 16.11
  

- OS: Windows 11

- Deployment Methods: Docker compose

- Docker image version: nocobase/nocobase:beta-full

- NodeJS version:

## * How To Reproduce

1.  Create a REST API datasource and one collection
2. Create a route and enter the URL into the browser with a search param containing the ID of one item from the collection: `http://localhost:13000/admin/<route id>?id=123456`
3. Add a Details block connected to the created collection and set the data scope to match the URL parameter.
4. The Details block sends its requests to the collection List action:
`request GET /api/<collection>:list?filter=%7B%22$and%22:[%7B%22$and%22:[%7B%22id%22:%7B%22$eq%22:%22123456%22%7D%7D]%7D]%7D&page=1&pageSize=1`
decoded as :
`GET /api/<collection>:list?filter={"$and":[{"$and":[{"id":{"$eq":"123456"}}]}]}&page=1&pageSize=1`

## Expected behavior

The Collection **Get** action should be called instead with the ID of the item.

## Screenshots

N/A

## Logs

N/A

hongboji:
It's as expected, just like when you create a multi-record detail block using a data table.

mrkmcnamee:
Perhaps this is just a usability question then. For example, in a Table block I can configure a column with "Enable click-to-open". This opens a drawer which contains the full details of the item clicked. The Table block is populated using the List request action and the drawer is populated using the Get request action.

How do I use "+ Add block" on a page so that it will be populated with the contents of the Get request action? 
This is what I assumed the Details block was for.

Cheers,
Kevin

hongboji:
The request for the single record details block opened in the pop-up window is "GET".

mrkmcnamee:
Yes, that's fine. The question is how to have the GET request action called from a block added to a page (not as a pop-up, drawer, etc.)?

mrkmcnamee:
[...]

PR #9086, opened 2026-04-12
fix: details block with pagination uses get action for non-DB datasources (REST API)

safsafa-dev:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Motivation
When using a REST API datasource, the Details block calls the `list` action instead of `get`. REST API `list` and `get` endpoints return different data structures, so the Details block was missing data that only the `get` endpoint provides.

### Description
In `useDetailsWithPaginationDecoratorProps`, detect if the datasource is a non-DB instance using `isDBInstance`. If so, switch to `action: 'get'` with `filterByTk` instead of `list` with a filter.

- DB datasources (main, external DB): unchanged, still uses `list` ✅
- Non-DB datasources (REST API): now uses `get` with `filterByTk` ✅

### Related issues
#8549

### Showcase
N/A

### Changelog
| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fixed details block calling `list` instead of `get` for REST API datasources |
| 🇨🇳 Chinese | 修复详情区块在 REST API 数据源中调用 `list` 而非 `get` 的问题 |

### Checklists

### DIFF
--- a/packages/core/client/src/modules/blocks/data-blocks/details-multi/hooks/useDetailsWithPaginationDecoratorProps.ts
+++ b/packages/core/client/src/modules/blocks/data-blocks/details-multi/hooks/useDetailsWithPaginationDecoratorProps.ts
@@ -6,19 +6,33 @@
  * This project is dual-licensed under AGPL-3.0 and NocoBase Commercial License.
  * For more information, please refer to: https://www.nocobase.com/agreement.
  */
-
 import { useParentRecordCommon } from '../../../useParentRecordCommon';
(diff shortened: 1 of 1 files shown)


safsafa-dev:
The CI failure seems to come from the "Deploy client docs" step, which requires SSH secrets.

The build steps (zh-CN and en-US) completed successfully, so the issue does not appear to be related to the changes in this PR.

This is expected for external PRs, since secrets are not available in GitHub Actions.

safsafa-dev:
Hello @Albert-mah 👋
I would appreciate your review when you have time.
Thank you so much!

PR #9103, opened 2026-04-13
fix(file-manager): improve preview download error handling

gaston98765:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Motivation

Improve reliability of file downloads from the preview modal.

### Description

This PR improves the robustness of preview file downloads by:

- checking `response.ok` before creating the blob
- adding error handling around the download flow
- removing unnecessary `new Blob([blob])`
- delaying `URL.revokeObjectURL` slightly to avoid revoking the object URL too early

These changes help prevent silent failures when downloading files, especially when files are served from external storage providers or CDNs.

### Related issues

N/A

### Showcase

N/A

### Changelog

| Language | Changelog |
|----------|----------|
| 🇺🇸 English | Improved preview file download error handling |
| 🇨🇳 Chinese | 改进了预览文件下载时的错误处理 |

### Docs

| Language | Link |
|----------|------|
| 🇺🇸 English | N/A |
| 🇨🇳 Chinese | N/A |

### Checklists

- [x] Doc update is not needed
- [x] Changelog is provided

### DIFF
--- a/packages/plugins/@nocobase/plugin-file-manager/src/client/models/DisplayPreviewFieldModel.tsx
+++ b/packages/plugins/@nocobase/plugin-file-manager/src/client/models/DisplayPreviewFieldModel.tsx
@@ -118,7 +118,7 @@ const Preview = (props) => {
     }
   }, [list.length, previewOpen]);
   const onDownload = React.useCallback(
-    (fileOverride?: any) => {
+    async (fileOverride?: any) => {
       const file = fileOverride || list[current];
       if (!file) {
         return;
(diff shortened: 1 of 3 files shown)


gaston98765:
Hi @mytharcher 👋
I would appreciate your review when you have time.
Thank you!

gaston98765:
Thanks for the suggestions 

I’ve updated the implementation to:
- use async/await with proper error handling
- move cleanup into a finally block
- replace the delay magic number with a named constant
- surface a UI error message when download fails

Let me know if anything else should be improved.

gaston98765:
Hi @mytharcher @hongboji ,

Just a quick follow-up on this PR,All suggestions have been addressed, and the checks are passing.

Thanks for your time.

packages/plugins/@nocobase/plugin-file-manager/src/client/models/DisplayPreviewFieldModel.tsx
```diff
@@ -117,35 +117,62 @@ const Preview = (props) => {
...
+
+        link = document.createElement('a');
+        link.href = blobUrl;
+        link.download = downloadName;
+
+        document.body.appendChild(link);
+        link.click();
+      } catch (error) {
+        console.error('File download failed:', error);
+        message.error('File download failed');
```
mytharcher: Please add this language text to locale/*.json as i18n content, and use translate function to parse.

mytharcher (changes requested):

packages/plugins/@nocobase/plugin-file-manager/src/client/models/DisplayPreviewFieldModel.tsx
gaston98765: Thanks for the feedback!

I’ve updated the implementation to use the i18n translation function and added the corresponding entries in the locale files (en-US and zh-CN).

Please let me know if anything else should be improved.

mytharcher (approved):

PR #9174, opened 2026-04-20
fix(collection-repository): prevent crash when fields is undefined

saraTabbane:
This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

## Motivation
Prevent potential runtime errors when loading collections if `fields` is undefined or not an array.

In `CollectionRepository.load`, the code uses `.filter()` and `.map()` directly on `fields`. If `fields` is missing or invalid, it can throw an error like:

> Cannot read properties of undefined (reading 'filter')

## Description
This PR adds defensive checks to ensure `fields` is a valid array before applying `.filter()` and `.map()` operations.

This change:
- adds `Array.isArray(fields)` guards in two places inside `CollectionRepository.load`
- returns an empty array when `fields` is not a valid array
- preserves existing behavior when `fields` is correctly defined

## Related issues
N/A

## Showcase
N/A

## Changelog
| Language | Changelog |
|----------|----------|
| 🇺🇸 English | Prevent crash when fields is undefined during collection loading |
| 🇨🇳 Chinese | 防止在集合加载期间 fields 未定义时发生崩溃 |

## Docs
| Language | Link |
|----------|------|
| 🇺🇸 English | |
| 🇨🇳 Chinese |  |

## Checklists
- [x] Doc update is not needed
- [x] Changelog is provided

### DIFF
--- a/packages/plugins/@nocobase/plugin-data-source-main/src/server/repositories/collection-repository.ts
+++ b/packages/plugins/@nocobase/plugin-data-source-main/src/server/repositories/collection-repository.ts
@@ -90,6 +90,10 @@ export class CollectionRepository extends Repository {
 
         const fields = nameMap[instanceName].get('fields');
 
+        if (!Array.isArray(fields)) {
+          return [];
+        }
+
         return fields
           .filter((field) => field['type'] === 'belongsTo' || field['type'] === 'belongsToMany')
           .map((field) => field.get('name'));
@@ -121,6 +125,10 @@ export class CollectionRepository extends Repository {
 
       const skipField = (() => {
         const fields = nameMap[viewCollectionName].get('fields');
+
+        if (!Array.isArray(fields)) {
+          return [];
+        }
 
         return fields
           .filter((field) => {


saraTabbane:
Hi @mytharcher,
I would appreciate you review if you have time
Thank you !

mytharcher:
LGTM. But I wonder what case will cause the error? Could you provide a test case to cover this?

saraTabbane:
Hi @mytharcher, thank you for getting back to me!
Here are some possible cases i could think of:

- this.find({ filter, appends: ['fields'] }) does not correctly resolve the appended fields relation.
- A collection record exists, but its related field metadata is incomplete or not properly associated.
- A plugin installation, migration, or schema synchronization creates or updates a collection before its fields are fully initialized.
- A view collection references another collection whose fields relation is missing or not loaded correctly.

In these cases, nameMap[collectionName].get('fields') can return undefined, and the previous code would crash when calling .filter().

The added Array.isArray(fields) guard treats missing or invalid fields as an empty array, while keeping the existing behavior unchanged when fields is correctly loaded.

gaston98765:
Hi @saraTabbane @mytharcher,

These seem like valid edge cases where fields might be missing or not loaded. The Array.isArray(fields) check is a simple way to avoid crashes and keep the same behavior.

For testing, we could mock a case where fields is undefined or not appended.

Nice improvement

PR #9322, opened 2026-05-02
feat(input-number, table-v2): locale-aware default number format and right-align numeric columns

ferraro:
## Why

NocoBase currently hard-codes `'0,0.00'` (US thousand-comma / dot-decimal) as the default number separator, and table cells render numbers left-aligned regardless of field type. For users with German, French, Spanish, etc. locales this means **every numeric column has to be configured manually** to match the user's locale, and accountants/operators see decimal points jumping around because numbers are not column-aligned.

Concrete user pain: a German-locale user sees `14,470.40` instead of `14.470,40` and `760.00` not aligned with `1,234.56` above it.

## What

Two minimal, surgical changes that improve defaults for the 95% case while keeping full backwards compatibility.

### 1. Locale-aware default number format

`packages/core/client/src/schema-component/antd/input-number/ReadPretty.tsx`

- Add `LOCALE_DEFAULT_SEPARATOR` map covering en / de / fr / es / it / nl / pt → format string
- `formatNumber()` accepts an optional `locale` and uses it as a fallback when `separator` is not explicitly set
- `ReadPretty` reads the active locale via `useTranslation()` and passes `i18n.language` down

**Backwards compatible:** any field that already sets `separator` via `SchemaSettingsNumberFormat` is **unaffected** — the explicit value still wins. Only the implicit default changes per user locale.

### 2. Auto right-align numeric columns

`packages/core/client/src/schema-component/antd/table-v2/Table.tsx`

- Auto-set `align: 'right'` for table columns whose field interface is `number`, `integer`, or `percent`
- Decimal-aligned values are easier to compare row-to-row (accounting / spreadsheet convention)
- Logic extracted into a pure helper `resolveColumnAlign` for testability

**Backwards compatible:** explicit `align` in `x-component-props` still wins, so per-column overrides keep working.

### 3. `'0 0,00'` separator: bug fix, not a breaking change

The pre-existing token `'0 0,00'` was internally inconsistent. The token's own name contains a comma as decimal separator (the same convention as `'0,0.00'` where the comma marks the thousands separator and the dot marks the decimal). However, the `separators` map rendered it as dot-decimal (`{ thousands: ' ', decimal: '.' }`), and the dropdown label mirrored that bug as `100 000.00`.

This PR aligns the implementation with the token's name:

- `separators['0 0,00']` → `{ thousands: ' ', decimal: ',' }`
- Dropdown label `100 000.00` → `100 000,00` (in `SchemaSettingsNumberFormat`, `numberFormat.tsx`, and `DisplayPercentFieldModel.tsx`)

**Why this is not a real breaking change:**

- A user who picked `'0 0,00'` from the dropdown was looking at the label `100 000.00` and *also* the token name showing comma-decimal — the two hints contradicted each other, so user intent was ambiguous. The realistic intent is *space-thousands + comma-decimal* (a typographic format used in formal European publications); the previous space-thousands + dot-decimal rendering matched no real-world locale convention.
- Anyone who genuinely wanted dot-decimal output would have picked `'0,0.00'` (label `100,000.00`) or `'0.00'` (label `100000.00`) — far more obvious choices for that intent.
- The new behavior makes the token meaning, the dropdown label, and the rendered output all consistent for the first time.

### 4. Tests

Updated/added tests in `input-number.test.tsx`:
- `en-US` → `1,234,567.89`
- `de-DE` → `1.234.567,89`
- `fr-FR` → `1 234 567,89` (now correctly comma-decimal)
- explicit separator overrides locale
- unknown locale falls back to en-US
- language-only code (`de` without region) is supported
- existing `'0 0,00'` separator test updated to assert comma-decimal output

New tests in `__tests__/resolveColumnAlign.test.ts`:
- numeric interfaces (`number` / `integer` / `percent`) default to `align: 'right'`
- non-numeric interfaces have no default align
- explicit `align` overrides the numeric default
- explicit `align` is preserved for non-numeric interfaces

## Out of scope

- No System Settings UI for global override (could come in a follow-up; the locale-mapping table is the simplest possible default)
- No changes to `SchemaSettingsNumberFormat` per-field UI (the dropdown options are unchanged; only the misleading `'0 0,00'` label is corrected)
- No changes to date/time format (separate concern)
- No locale-aware currency symbol placement (currency is currently rendered via manual `addonBefore`/`addonAfter` per field; can come in a follow-up)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/client/src/schema-component/antd/input-number/ReadPretty.tsx
+++ b/packages/core/client/src/schema-component/antd/input-number/ReadPretty.tsx
@@ -58,7 +58,7 @@ function countDecimalPlaces(value) {
 const separators = {
   '0,0.00': { thousands: ',', decimal: '.' },
   '0.0,00': { thousands: '.', decimal: ',' },
-  '0 0,00': { thousands: ' ', decimal: '.' },
+  '0 0,00': { thousands: ' ', decimal: ',' },
   '0.00': { thousands: '', decimal: '.' }, // 没有千位分隔符
 };
 //分隔符换算
--- a/packages/core/client/src/schema-component/antd/input-number/ReadPretty.tsx
+++ b/packages/core/client/src/schema-component/antd/input-number/ReadPretty.tsx
@@ -13,8 +13,36 @@ import BigNumber from 'bignumber.js';
 import { format } from 'd3-format';
 import * as math from 'mathjs';
 import React, { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
 import { toString } from 'lodash';
 import { withPopupWrapper } from '../../common/withPopupWrapper';
+
+// Locale-aware default separator. Most non-English European locales use
+// dot-thousands and comma-decimal; English uses comma-thousands and dot-decimal.
+// The literal '0,0.00' format string is interpreted by `separators` below — it is
+// *not* a literal output, so the mapping is correct.
+const LOCALE_DEFAULT_SEPARATOR: Record<string, '0,0.00' | '0.0,00' | '0 0,00' | '0.00'> = {
+  'en-US': '0,0.00',
+  'en-GB': '0,0.00',
+  en: '0,0.00',
+  'de-DE': '0.0,00',
+  de: '0.0,00',
+  'es-ES': '0.0,00',
+  es: '0.0,00',
+  'it-IT': '0.0,00',
+  it: '0.0,00',
+  'nl-NL': '0.0,00',
+  nl: '0.0,00',
+  'fr-FR': '0 0,00',
+  fr: '0 0,00',
+  'pt-BR': '0.0,00',
+  pt: '0.0,00',
+};
+
+function getLocaleDefaultSeparator(locale?: string): '0,0.00' | '0.0,00' | '0 0,00' | '0.00' {
+  if (!locale) return '0,0.00';
+  return LOCALE_DEFAULT_SEPARATOR[locale] || LOCALE_DEFAULT_SEPARATOR[locale.split('-')[0]] || '0,0.00';
+}
 
 function countDecimalPlaces(value) {
   const strValue = toString(value);
@@ -134,7 +162,11 @@ export function scientificNotation(number, decimalPlaces, separator = '.') {
 }
 
 export function formatNumber(props) {
-  const { step, formatStyle = 'normal', value, unitConversion, unitConversionType, separator = '0,0.00' } = props;
+  const { step, formatStyle = 'normal', value, unitConversion, unitConversionType, separator, locale } = props;
+  // Backwards-compatible: explicit `separator` (set by SchemaSettingsNumberFormat)
+  // wins. If undefined, fall back to a locale-aware default so e.g. `de-DE`
+  // sessions render `1.234,56` and `en-US` renders `1,234.56` out of the box.
+  const effectiveSeparator = separator ?? getLocaleDefaultSeparator(locale);
 
   if (!isValid(value)) {
     return null;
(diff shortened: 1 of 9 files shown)


packages/core/client/src/schema-component/antd/table-v2/Table.tsx
```diff
@@ -238,6 +248,14 @@ const useTableColumns = (
           uiSchema.default = defaultValue;
         }
 
+        // Auto right-align numeric columns (decimal-aligned reads better in tables;
+        // matches accounting/spreadsheet convention). Explicit `align` in
+        // x-component-props still wins.
+        const explicitAlign = columnSchema['x-component-props']?.align;
```
zhangzhonghe: This default right-alignment change makes sense to me, but I think it would be good to add a focused regression test before merging. The current TableV2 tests are mostly smoke renders, so they do not really protect the two important contracts introduced here: numeric interfaces (`number` / `integer` / `percent`) default to `align: right`, and an explicit `x-component-props.align` still overrides that default. Since this is user-visible behavior, a small targeted test would make the change much safer.

packages/core/client/src/schema-component/antd/input-number/ReadPretty.tsx
```diff
@@ -13,9 +13,37 @@ import BigNumber from 'bignumber.js';
...
+  en: '0,0.00',
+  'de-DE': '0.0,00',
+  de: '0.0,00',
+  'es-ES': '0.0,00',
+  es: '0.0,00',
+  'it-IT': '0.0,00',
+  it: '0.0,00',
+  'nl-NL': '0.0,00',
+  nl: '0.0,00',
+  'fr-FR': '0 0,00',
```
zhangzhonghe, line 36: Nice improvement overall. One thing to watch here: `fr-FR` / `fr` is currently mapped to `0 0,00`, but in this module that existing token still means space thousands + dot decimal (`100 000.00`), not a French-style comma decimal. That means French users would still get `1 234 567.89`, so the locale-aware behavior is not quite correct yet. I do not think we should repurpose `0 0,00`, because that could change the meaning of already-saved field configs. A safer fix would be to introduce a new separator token for space thousands + comma decimal, map `fr*` to that new token, and update the related tests/settings labels as well.

ferraro: Thanks for the careful review. I'd like to argue this is best understood as a bug fix rather than a behavior change, and I've pushed a commit that adopts that framing.

The token `'0 0,00'` already names comma-decimal in its own string. Every token in this map follows the same convention: in `'0,0.00'` the comma is the thousands separator and the dot is the decimal; in `'0.0,00'` the dot is thousands and the comma is decimal. By that consistent reading, `'0 0,00'` has always meant *space thousands, comma decimal*. The previous implementation (`{ thousands: ' ', decimal: '.' }`) silently contradicted the token's own name and produced `100 000.00` — a format that doesn't match any real-world locale convention I'm aware of.

The dropdown label `100 000.00` mirrored the buggy behavior, but a user picking this token saw two contradictory hints (label says dot decimal, token name says comma decimal). Realistically:

- Anyone wanting unambiguous dot-decimal output would pick `'0,0.00'` (label `100,000.00`) or `'0.00'` (label `100000.00`).
- The only coherent intent for picking `'0 0,00'` is space-thousands + comma-decimal — exactly what the token's name asks for.

If a user has saved this token in their schema, the contract they implicitly accepted is "the format my token name describes". Producing dot-decimal for a token whose name asks for comma-decimal was the bug; producing comma-decimal is honoring the contract.

I've aligned the implementation, the three dropdown labels (`SchemaSettingsNumberFormat.tsx`, `numberFormat.tsx`, `DisplayPercentFieldModel.tsx`), the existing `'0 0,00'` test, and the `fr-FR` locale default — all in commit 7401312271. There's also an explanation in the PR body under "`'0 0,00'` separator: bug fix, not a breaking change". Happy to keep discussing if you still see a real-world break case.

packages/core/client/src/schema-component/antd/table-v2/Table.tsx
ferraro: Good call — added in commit 7401312271. I extracted the auto-align logic into a small pure helper `resolveColumnAlign(explicitAlign, fieldInterface)` in `packages/core/client/src/schema-component/antd/table-v2/resolveColumnAlign.ts`, so the contract can be tested directly without spinning up a full table render.

New test file: `packages/core/client/src/schema-component/antd/table-v2/__tests__/resolveColumnAlign.test.ts`, covering:

- numeric interfaces (`number` / `integer` / `percent`) default to `align: 'right'`
- non-numeric interfaces have no default align (returns `undefined`)
- explicit `align` overrides the numeric default (`left` / `center` / `right`)
- explicit `align` is preserved for non-numeric interfaces

All 4 tests pass locally.

PR #9903, opened 2026-06-25
fix(plugin-workflow): address v2 workflow regressions

Molunerfinn:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Motivation
- Several v2 workflow configuration surfaces still diverge from the online v1 behavior.
- The first fix in this branch restores collection trigger condition inputs so the right-hand value editor stays type-aware but does not expose workflow variables.

### Description 
- add a right-side variable toggle to the shared workflow filter dynamic component while keeping variable input enabled by default
- disable right-side variable input only for collection trigger condition filters to match the v1 data table event trigger behavior
- keep condition-node calculation inputs unchanged and add regression coverage for both the shared filter component and collection trigger config

### Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### Docs

| Language   | Link |
| ---------- | --------- |
| 🇺🇸 English |  |
| 🇨🇳 Chinese |  |

### Checklists
- [ ] Changelog is provided or not needed

### DIFF
--- a/packages/core/client-v2/src/components/form/TypedVariableInput.tsx
+++ b/packages/core/client-v2/src/components/form/TypedVariableInput.tsx
@@ -530,12 +530,14 @@ export function TypedVariableInput(props: TypedVariableInputProps) {
   );
 
   const onClearVariable = useCallback(() => {
-    if (nullable) {
-      onChange?.(null);
+    const first = normalizedTypes[0];
+    if (first) {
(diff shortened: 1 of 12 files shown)


Molunerfinn:
self test done

PR #9950, opened 2026-06-30
fix(client): correct Japanese translation of "Tag" (currently shows "Tab")

mahirhir:
In `locales/ja-JP.json`, the `Tag` string under `@nocobase/client` is translated as `タブ`, which is the Japanese word for "Tab". The Japanese word for "Tag" is `タグ`.

The same namespace also has a separate `Tab` key whose value is `タブ`, so right now both "Tag" and "Tab" show up as `タブ` in Japanese. That makes the tag label read as "Tab" to Japanese users.

The other locales keep the two distinct: Korean uses `태그` for Tag and `탭` for Tab, and Chinese uses `标签` for Tag. Only Japanese collapsed them into the same word.

This fixes the single value:

```
"Tag": "タブ"  ->  "Tag": "タグ"
```

### DIFF
--- a/locales/ja-JP.json
+++ b/locales/ja-JP.json
@@ -1319,7 +1319,7 @@
     "Table size": "テーブルサイズ",
     "Tablet": "Tablet",
     "Tablet device": "タブレットデバイス",
-    "Tag": "タブ",
+    "Tag": "タグ",
     "Tag color field": "ラベルの色フィールド",
     "Target": "ターゲット",
     "Target block UID": "Target block UID",


mahirhir:
This PR is 1221 commits behind `main`, so it looks stale. The string is not: still wrong on `main` today, and the measurement is small enough to check in one paste.

```console
$ curl -s .../nocobase/main/locales/ja-JP.json | grep -n '"Tab"\|"Tag"'
1311:    "Tab": "タブ",
1322:    "Tag": "タブ",      # ← should be タグ

$ grep -c '": "タブ"' ja-JP.json     # keys rendering as タブ
2
$ grep -c '": "タグ"' ja-JP.json     # keys rendering as タグ
0
```

Two different English terms render as the same Japanese word, and **タグ does not appear anywhere in the Japanese locale at all**. A Japanese user has no way to tell a tag from a tab in the UI, and never sees the word for "tag".

The distinction is not subtle in Japanese. タブ is the tab of a tabbed interface; タグ is a label you attach to a record. They are unrelated concepts that happen to be one katakana character apart, which is presumably how it got in.

The change is that one character. No behaviour, no schema, no other locale touched:

```diff
-    "Tag": "タブ",
+    "Tag": "タグ",
```

CI is green and the branch is `MERGEABLE` despite the drift, since a single JSON value does not conflict with anything. Happy to rebase if you would rather have it on top of current `main`, or close this and change the character directly; either is fine and needs no attribution.

PR #10088, opened 2026-07-14
fix(plugin-workflow): handle unresolved filter variables

Molunerfinn:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Motivation

Workflow collection filters previously converted workflow variable expressions into the FlowEngine `ctx` format. When a valid expression referenced a path missing from the returned metadata tree, it fell back to a raw input instead of displaying the parsing failure state.

### Description

- Allow `VariableFilterItem` to accept domain-specific right-side variable converters while preserving constant, null, and core fallbacks.
- Keep workflow filter expressions in the native `{{$...}}` format and support parsing both workflow and FlowEngine formats.
- Pass converter-resolved paths to `VariableTag` so unresolved metadata displays the red parsing failure pill.
- Add regression coverage for valid variables, unresolved paths, converter composition, and workflow filter round trips.

### Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Workflow filters now preserve workflow variable expressions and show a clear parsing error when variable metadata cannot be resolved. |
| 🇨🇳 Chinese | 工作流筛选条件现在会保留工作流变量表达式，并在变量元数据无法解析时显示明确的错误提示。 |

### Docs

| Language   | Link |
| ---------- | --------- |
| 🇺🇸 English | |
| 🇨🇳 Chinese | |

### Checklists

### DIFF
--- a/packages/core/client-v2/src/flow/components/filter/VariableFilterItem.tsx
+++ b/packages/core/client-v2/src/flow/components/filter/VariableFilterItem.tsx
@@ -131,6 +131,10 @@ export interface VariableFilterItemProps {
    * 默认使用整棵 ctx 的 metaTree：model.context.getPropertyMetaTree()
    */
   rightMetaTree?: MetaTreeNode[] | (() => MetaTreeNode[] | Promise<MetaTreeNode[]>);
+  /**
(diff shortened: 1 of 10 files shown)


Molunerfinn:
img_v3_0213j_0cbe444d-c695-4fc7-8c9a-ffe594c4d33g

PR #10326, opened 2026-08-10
fix(client-v2): preserve large decimal values

katherinehhh:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Motivation
V2 decimal fields convert values larger than the JavaScript safe integer range to scientific notation in the editor. After saving, the number display formatter treats that scientific-notation string as a regular decimal string and may render only the leading digit, even though the database and API retain the complete value.

### Description
- Serialize large editable number values with `BigNumber#toFixed()` so the editor keeps a regular decimal string.
- Use the same non-exponential serialization before formatting large values for display.
- Preserve the existing `number` value type for regular inputs.
- Add regression coverage for regular inputs, the reported 22-digit boundary value, and a 30-digit decimal value.

Risk is low: production behavior changes in two lines and only for values outside the JavaScript safe integer range. Existing number-format options and normal-value behavior remain unchanged. Explicitly configured unit conversion keeps its existing JavaScript-number precision behavior and is outside this fix.

Tested with:
- `yarn test packages/core/client-v2/src/flow/models/fields/__tests__/NumberFieldModel.test.tsx --run --reporter=verbose`
- `yarn test packages/core/client-v2/src/flow/models/fields/__tests__/DisplayNumberFieldModel.test.ts --run --reporter=verbose`
- `yarn eslint --fix` on all four changed files

Suggested manual verification: enter `123456789012345678901234567890` in a V2 `NUMERIC(38,0)` field, save the record, and confirm both the editor and table display retain the complete value.

### Related issues
- Task 5860: https://test_management.v2.test.nocobase.com/nocobase/admin/kbruj97dzuv/view/09d5fa415ca/filterbytk/543/view/a1989b2f48d/filterbytk/5860

### Showcase
Before: the editor changes the value to scientific notation and the table may display only `1`.

After: the editor retains the complete decimal value and the table displays `123,456,789,012,345,678,901,234,567,890`.

### Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix incorrect editing and display of large decimal values in V2 fields. |
| 🇨🇳 Chinese | 修复 V2 字段编辑和显示超大 decimal 数值不正确的问题。 |

### Docs

| Language   | Link |
| ---------- | --------- |
| 🇺🇸 English | Not needed |
| 🇨🇳 Chinese | 不需要 |

### Checklists

### DIFF
--- a/packages/core/client-v2/src/flow/models/fields/DisplayNumberFieldModel.tsx
+++ b/packages/core/client-v2/src/flow/models/fields/DisplayNumberFieldModel.tsx
@@ -71,7 +71,7 @@ function formatBigNumberWithSeparator(value, format = '0.00', step = 1, formatSt
   let number = value;
 
   if (formatStyle) {
-    number = new BigNumber(value).toString();
+    number = new BigNumber(value).toFixed();
   }
 
   let formattedNumber = '';
--- a/packages/core/client-v2/src/flow/models/fields/NumberFieldModel.tsx
+++ b/packages/core/client-v2/src/flow/models/fields/NumberFieldModel.tsx
@@ -20,7 +20,7 @@ function toSafeNumber(value) {
     return value;
   }
   if (value > Number.MAX_SAFE_INTEGER || value < Number.MIN_SAFE_INTEGER) {
-    return new BigNumber(value).toString();
+    return new BigNumber(value).toFixed();
   } else {
     return Number(value);
   }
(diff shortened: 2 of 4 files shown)


katherinehhh:
Superseded by #10330; the target branch has been changed from main to next.

PR #10401, opened 2026-08-22
fix: answer 404 for requests against a destroyed collection

### LINKED ISSUE #10397: A data request to a destroyed collection answers 500 TypeError, where the same API answers 404 for a collection that never existed
## * Describe the bug

After a collection is removed with `collections:destroy`, a data request against its name does not answer `404`. It answers `500` with a JavaScript type error:

```
GET /api/tcdestroy:list
-> 500 {"errors":[{"message":"Cannot read properties of undefined (reading 'collection')"}]}
```

The destroy itself is correct. The table really is gone from PostgreSQL. What is wrong is the answer the API gives afterwards: a caller that reads a collection which has since been dropped gets an internal error instead of "this resource is not here", and any client that distinguishes 4xx from 5xx will treat a normal, expected condition as a server fault, retry it, and page someone.

#### The control, which is the argument

This is not a claim about what `404` *should* mean here. It is what this very API already answers for the only other way a collection can be absent:

```
GET /api/tcneverexisted:list
-> 404 Not Found
```

Same endpoint, same shape of question, two different answers. A name that never existed is a clean `404`. A name that existed until a moment ago is a `500` type error. The product already has the right answer for an absent collection; the destroyed case simply does not reach it.

#### Cause

The resourcer route outlives the collection, so the request is dispatched normally and the action runs with no repository behind it.

`packages/core/actions/src/utils.ts:26-39`, `getRepositoryFromParams(ctx)` ends at

```ts
return ctx.db.getRepository<Repository>(resourceName);
```

which is `undefined` once the collection is gone. There is no undefined guard on the way out.

`packages/core/actions/src/actions/list.ts:37`:

```ts
let { simplePaginate } = repository.collection?.options || {};
```

https://github.com/nocobase/nocobase/blob/032a4f6913be912f57462d605cbd0bde97b599c6/packages/core/actions/src/actions/list.ts#L37

The optional chain sits one level too deep. It guards `.options`, but `repository` itself is the undefined value, so reading `.collection` off it throws `TypeError: Cannot read properties of undefined (reading 'collection')`. The live server's own stack agrees: the throw is inside `listWithPagination` at `.../actions/list.js:66:46`.

## * Environment

Stock install, no extra plugins enabled.

## * How To Reproduce

Everything below is verbatim from one run on an instance built from the commit above. All values are the real ones from that run.

**1. Create a collection, give it a field, and put a row in it.** As the super admin:

```
POST /api/collections:create
{"name":"tcdestroy","title":"TC Destroy","template":"general"}
-> 200

POST /api/collections/tcdestroy/fields:create
{"name":"label","type":"string","interface":"input"}
-> 200

POST /api/tcdestroy:create
{"label":"row one"}
-> 200 {"data":{"id":1,"label":"row one"}}
```

**2. Confirm the collection reads normally before the destroy.** This is what makes the later 500 a change of behaviour rather than a broken fixture:

```
GET /api/tcdestroy:list
-> 200 {"data":[{"id":1,"label":"row one"}],"meta":{"count":1,"page":1,"pageSize":20,"totalPage":1}}
```

**3. Destroy it.**

```
POST /api/collections:destroy?filterByTk=tcdestroy
-> 200 {"data":1}
```

The destroy is correct and complete. Read straight from PostgreSQL:

```
select count(*) from information_schema.tables where table_name='tcdestroy';
-> 0
```

**4. Read the destroyed collection. This is the bug.**

```
GET /api/tcdestroy:list
-> 500 {"errors":[{"message":"Cannot read properties of undefined (reading 'collection')"}]}
```

**5. The control, a name that never existed at all.** Nothing was created under this name at any point in the session:

```
GET /api/tcneverexisted:list
-> 404 Not Found
```

## Expected behavior

`GET /api/tcdestroy:list` after the destroy should answer `404`, the same as `GET /api/tcneverexisted:list`, since from the caller's point of view both describe a collection that is not there.

Two ways to get there, in preference order:

1. **Fix it at the source.** Have `getRepositoryFromParams` (`packages/core/actions/src/utils.ts:26-39`) throw a 404 when the named resource has no repository. That is one place, and it covers every action that goes through it rather than just `list`. We drove only `list` in this report, so we are not claiming the other actions are broken in the same way, but they take the same helper and are worth checking in the same change.

2. **Or at least stop the type error.** Guard `repository` in `listWithPagination` before dereferencing it, so the failure is an error the API chose rather than a crash. This removes the 500 but leaves each action to answer for itself.

Either way the caller should be able to tell a missing resource from a server fault, which today it cannot.

## Screenshots

Not applicable. The finding is entirely in the HTTP responses quoted above, and the decisive pair (500 for the destroyed name, 404 for the never-created one) is inline.

## Logs

The response body is the whole log:

```
{"errors":[{"message":"Cannot read properties of undefined (reading 'collection')"}]}
```

The server-side stack frame for the same request is `listWithPagination` at `.../packages/core/actions/lib/actions/list.js:66:46`, which is the compiled form of `packages/core/actions/src/actions/list.ts:37` quoted above.

#### Suggested labels

`bug` (our account cannot apply labels itself).

#### Related

We searched the tracker before filing, including the phrase `reading 'collection'`. Eight older items contain that phrase and none of them is this bug: they are upgrade-time crashes, a data-source block placeholder fix, a view sync error and an export action. If there is an existing report we missed, please close this as a duplicate.

Found by TrueCourse running the product's published documentation against a live instance; the full transcript (requests, responses, and the PostgreSQL readings taken after the destroy) is available on request.

yzxcj797:
## What?

After `collections:destroy`, a data request against the collection's name answers `500` with `TypeError: Cannot read properties of undefined (reading 'collection')` — while the same endpoint answers a clean `404` for a collection that never existed.

Closes #10397

## Why?

The resourcer route outlives the collection: the request still reaches the action after the table is dropped, and `getRepositoryFromParams` ends at

```ts
return ctx.db.getRepository<Repository>(resourceName);
```

which is `undefined` once the collection is gone — there is no guard on the way out. The action then dereferences the repository (`repository.collection?.options` in `list.ts` — the optional chain sits one level too deep, as the report notes) and the TypeError surfaces as a 500. A caller that distinguishes 4xx from 5xx treats a normal, expected condition (a collection that existed until a moment ago) as a server fault, retries it, and pages someone.

## How?

Guard the single exit point of `getRepositoryFromParams`: all three resolution branches (plain resource, `sourceId` association, and the `_` get/list special case, which previously crashed one line earlier on `collection.name`) now funnel into one check — when nothing resolves, `ctx.throw(404, ...)` naming the collection, matching the API's existing answer for absent collections. No behavior changes for any request that resolves a repository.

## Testing

New `destroyed-collection-404.test.ts` (same `mockServer` harness as the other action suites):

- list/get/create against a removed collection answer **404**;
- a collection that never existed still answers 404.

**Differential**: on `main`, the destroyed-collection list case fails with the exact error from the report — `TypeError: Cannot read properties of undefined (reading 'collection')` — and with this change all four cases pass. The existing action suites (`list`, `get`, `create`, `update`, `destroy`) pass unchanged (list has 2 pre-existing skips).

### DIFF
--- a/packages/core/actions/src/__tests__/destroyed-collection-404.test.ts
+++ b/packages/core/actions/src/__tests__/destroyed-collection-404.test.ts
@@ -1,0 +1,63 @@
+/**
+ * This file is part of the NocoBase (R) project.
+ * Copyright (c) 2020-2024 NocoBase Co., Ltd.
+ * Authors: NocoBase Team.
+ *
+ * This project is dual-licensed under AGPL-3.0 and NocoBase Commercial License.
+ * For more information, please refer to: https://www.nocobase.com/agreement.
+ */
+
+import { registerActions } from '@nocobase/actions';
+import { mockServer, MockServer } from './index';
+
+describe('destroyed collection', () => {
+  let app: MockServer;
+
+  beforeEach(async () => {
+    app = mockServer();
+    registerActions(app);
+
+    app.collection({
+      name: 'tcdestroy',
+      fields: [{ type: 'string', name: 'name' }],
+    });
+
+    await app.db.sync();
+  });
+
+  afterEach(async () => {
+    await app.destroy();
+  });
+
+  it('answers 404 for a list request once the collection is destroyed', async () => {
+    const before = await app.agent().resource('tcdestroy').list();
+    expect(before.status).toBe(200);
+
+    app.db.removeCollection('tcdestroy');
+
+    const response = await app.agent().resource('tcdestroy').list();
+    expect(response.status).toBe(404);
+  });
+
+  it('answers 404 for a get request once the collection is destroyed', async () => {
+    const before = await app.agent().resource('tcdestroy').list();
+    expect(before.status).toBe(200);
+
+    app.db.removeCollection('tcdestroy');
+
+    const response = await app.agent().resource('tcdestroy').get({ filterByTk: 1 });
+    expect(response.status).toBe(404);
+  });
+
+  it('answers 404 for a create request once the collection is destroyed', async () => {
+    app.db.removeCollection('tcdestroy');
+
+    const response = await app.agent().resource('tcdestroy').create({ values: { name: 'x' } });
+    expect(response.status).toBe(404);
+  });
+
+  it('still answers 404 for a collection that never existed', async () => {
+    const response = await app.agent().resource('tcneverexisted').list();
+    expect(response.status).toBe(404);
+  });
+});
--- a/packages/core/actions/src/utils.ts
+++ b/packages/core/actions/src/utils.ts
@@ -26,16 +26,28 @@ export function pageArgsToLimitArgs(
 export function getRepositoryFromParams(ctx: Context) {
   const { resourceName, sourceId, actionName } = ctx.action;
 
+  let repository: Repository | MultipleRelationRepository | undefined;
+
   if (sourceId === '_' && ['get', 'list'].includes(actionName)) {
     const collection = ctx.db.getCollection(resourceName);
-    return ctx.db.getRepository<Repository>(collection.name);
+    repository = collection ? ctx.db.getRepository<Repository>(collection.name) : undefined;
+  } else if (sourceId) {
+    repository = ctx.db.getRepository<MultipleRelationRepository>(resourceName, sourceId);
+  } else {
+    repository = ctx.db.getRepository<Repository>(resourceName);
   }
 
-  if (sourceId) {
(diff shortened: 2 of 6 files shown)


truecourse-agent:
I reported #10397, where a data request against a destroyed collection answered 500 with "Cannot read properties of undefined" instead of a 404.

✅ Re-ran that case on this branch and it now answers 404. The whole flow passes.

PR #10402, opened 2026-08-22
fix: stop the default-mode __union__ substitution from being silent

### LINKED ISSUE #10399: Under the default role mode, X-Role: __union__ is silently rewritten to the user's first role instead of being refused like every other unheld role name
## * Describe the bug

The ACL middleware refuses every role name a user does not hold, with a `401` and a specific code:

```
GET /api/tcunionmixed:list?sort=id   X-Role: tcnosuchrole
-> 401 {"errors":[{"message":"The role does not belong to the user","code":"ROLE_NOT_FOUND_FOR_USER"}]}
```

There is exactly one exception. Under the **default** role mode, `X-Role: __union__` is not refused. It is quietly rewritten to `userRoles[0].name`, and the request runs as that role and answers `200`. The response carries nothing to say a substitution happened, so a caller who asked to act as the union of their roles is told, in every observable way, that it worked.

**This is not a privilege escalation report and should not be read as one.** `userRoles[0]` is always a role the caller already holds, so the effective permissions of the substituted request are a subset of what the caller asked for, never a superset. Nothing is leaked and no boundary is crossed. The defect is the silent substitution itself: the same middleware, four lines further down, refuses every other name it cannot honour, and the caller has no way to learn that the answer in their hands came from a different role than the one they named.

The practical cost is that a client cannot tell "the union" from "one of your roles". Under the default mode a script that sends `__union__` receives one role's rows and treats them as the union's rows, and no error is raised at any layer.

#### Three controls, and each removes a different objection

**Control 1: the answer is byte-identical to naming the first role explicitly.**

```
GET /api/tcunionmixed:list?sort=id   X-Role: tcuniona
-> 200 {"data":[{"age":23,"name":"Jack","id":1},{"age":29,"name":"Lily","id":2},{"age":27,"name":"Jade","id":3}],"meta":{"count":3,"page":1,"pageSize":20,"totalPage":1}}

GET /api/tcunionmixed:list?sort=id   X-Role: __union__
-> 200 {"data":[{"age":23,"name":"Jack","id":1},{"age":29,"name":"Lily","id":2},{"age":27,"name":"Jade","id":3}],"meta":{"count":3,"page":1,"pageSize":20,"totalPage":1}}
```

Compared as raw text, the two response bodies are equal.

**Control 2: the server itself reports the substituted role.**

```
GET /api/roles:check   X-Role: __union__   -> data.role = "tcuniona"
GET /api/roles:check   X-Role: tcuniona    -> data.role = "tcuniona"
```

So this is not an inference from the row set. The product's own endpoint for "which role am I acting as" answers `tcuniona` for a request that named `__union__`.

**Control 3: a real union looks different, so the default-mode answer was not a union that happened to match.** Flip the system role mode and re-issue the same request with the same header:

```
POST /api/roles:setSystemRoleMode {"roleMode":"allow-use-union"}  -> 200

GET /api/tcunionmixed:list?sort=id   X-Role: __union__
-> 200 {"data":[{"name":"Jack","sex":"Man","id":1,"age":23},
                {"name":"Lily","sex":"Woman","id":2,"age":29},
                {"name":"Jade","sex":"Woman","id":3,"age":27},
                {"name":"James","sex":"Man","id":4,"age":31}],"meta":{"count":4,...}}
```

Four rows, and both `age` and `sex`. The default-mode answer had three rows and no `sex`, so it was one role's answer and not a union. This is the control that matters, because without it a reader can reasonably say the 200 *was* the union.

#### Cause

`packages/plugins/@nocobase/plugin-acl/src/server/middlewares/setCurrentRole.ts:55-58`:

```ts
if ([currentRole, ctx.state.currentRole].includes(UNION_ROLE_KEY) && roleMode === SystemRoleMode.default) {
  currentRole = userRoles[0].name;
  ctx.state.currentRole = userRoles[0].name;
  ctx.headers['x-role'] = userRoles[0].name;
}
```

https://github.com/nocobase/nocobase/blob/032a4f6913be912f57462d605cbd0bde97b599c6/packages/plugins/@nocobase/plugin-acl/src/server/middlewares/setCurrentRole.ts#L55-L58

Note that the incoming request header is rewritten too, so downstream code cannot see what the caller actually sent.

Every other unheld name falls through to `:74-81` and is thrown out:

```ts
role = userRoles.find((role) => role.name === currentRole)?.name;
if (!role) {
  return ctx.throw(401, { code: 'ROLE_NOT_FOUND_FOR_USER', ... });
}
```

`UNION_ROLE_KEY` is `'__union__'` (`packages/plugins/@nocobase/plugin-acl/src/server/constants.ts:10`). The two branches sit in the same file, twenty lines apart, and treat the same class of input in opposite ways.

## * Environment

Stock install. The system role mode was the installed default: `select "roleMode" from "systemSettings"` returns `default`.

## * How To Reproduce

All values below are verbatim from one run on an instance built from the commit above.

**1. A collection with four rows.** Collection `tcunionmixed`, columns `name`, `age`, `sex`:

```
1|Jack|23|Man
2|Lily|29|Woman
3|Jade|27|Woman
4|James|31|Man
```

**2. Two roles, each with its own row filter and column set.**

- `tcuniona`: `view` scoped `{"$and":[{"age":{"$lt":30}}]}`, fields `["name","age"]`
- `tcunionb`: `view` scoped `{"$and":[{"name":{"$includes":"Ja"}}]}`, fields `["name","sex"]`

**3. One user holding exactly those two roles.** User `tcunion`; checked in the database so the result cannot be blamed on a third implicit role, `rolesUsers` holds `tcuniona` and `tcunionb` for this user and nothing else. Sign in as this user for every request below.

**4. Confirm the mode.**

```
select "roleMode" from "systemSettings";
-> default
```

**5. Read as the first role, then read as the union. This is the bug.**

```
GET /api/tcunionmixed:list?sort=id   X-Role: tcuniona
-> 200 {"data":[{"age":23,"name":"Jack","id":1},{"age":29,"name":"Lily","id":2},{"age":27,"name":"Jade","id":3}],"meta":{"count":3,"page":1,"pageSize":20,"totalPage":1}}

GET /api/tcunionmixed:list?sort=id   X-Role: __union__
-> 200 {"data":[{"age":23,"name":"Jack","id":1},{"age":29,"name":"Lily","id":2},{"age":27,"name":"Jade","id":3}],"meta":{"count":3,"page":1,"pageSize":20,"totalPage":1}}
```

The two raw response texts compare equal.

**6. Ask the server which role it used.**

```
GET /api/roles:check   X-Role: __union__   -> data.role = "tcuniona"
GET /api/roles:check   X-Role: tcuniona    -> data.role = "tcuniona"
```

**7. Control, every other unheld name.** `admin` exists but this user does not hold it; `tcnosuchrole` does not exist at all. Both are refused:

```
GET /api/tcunionmixed:list?sort=id   X-Role: admin
-> 401 {"errors":[{"message":"The role does not belong to the user","code":"ROLE_NOT_FOUND_FOR_USER"}]}

GET /api/tcunionmixed:list?sort=id   X-Role: tcnosuchrole
-> 401 {"errors":[{"message":"The role does not belong to the user","code":"ROLE_NOT_FOUND_FOR_USER"}]}
```

`__union__` is the only unheld name that does not get this treatment.

**8. Control, what a genuine union answers.** Flip the mode, verify it in the database, and send the same request with the same header:

```
POST /api/roles:setSystemRoleMode {"roleMode":"allow-use-union"}   -> 200
select "roleMode" from "systemSettings";   -> allow-use-union

GET /api/tcunionmixed:list?sort=id   X-Role: __union__
-> 200 {"data":[{"name":"Jack","sex":"Man","id":1,"age":23},
                {"name":"Lily","sex":"Woman","id":2,"age":29},
                {"name":"Jade","sex":"Woman","id":3,"age":27},
                {"name":"James","sex":"Man","id":4,"age":31}],"meta":{"count":4,...}}
```

Four rows and an extra column, from the same request that returned three rows and no `sex` a moment earlier.

## Expected behavior

`X-Role: __union__` under the default role mode should be answered the way every other role the caller cannot use is answered:

```
401 {"errors":[{"message":"The role does not belong to the user","code":"ROLE_NOT_FOUND_FOR_USER"}]}
```

or a `400` with a code of its own, for example that the union role is not available under the current system role mode. Either is fine. What matters is that the caller learns the request was not honoured as asked, instead of receiving a `200` produced under a role it never named.

If degrading to the first role is a deliberate compatibility behaviour, then two things should follow. The response should carry a signal that the substitution happened, so a client can react to it, and the role-mode documentation should state plainly that under the default mode `__union__` resolves to the user's first role rather than to a union. Today neither is true, and the request header is rewritten in place at `setCurrentRole.ts:57`, so even server-side code downstream cannot tell what the caller sent.

One more note on scope, because it decides how this should be triaged: this is a correctness and observability bug, not a security one. The substituted role is one the caller holds, and its permissions are a subset of the union that was requested.

## Screenshots

Not applicable. The finding is entirely in the HTTP responses quoted above, together with the two `roleMode` readings taken from PostgreSQL.

## Logs

There is no error log to attach, and that is part of the report. The request never reaches an error path: the middleware rewrites the role and the action then runs normally, so nothing is logged as a failure at any layer. The record of what happened is the pair of byte-identical response bodies, the `roles:check` answer naming `tcuniona`, and the different result the same header produces under `allow-use-union`, all quoted above.

#### Suggested labels

`bug`, `users & permissions` (our account cannot apply labels itself).

#### Related

We searched the tracker for `__union__` and for recently created items before filing. The union-related pull requests we found are unrelated code fixes that predate the commit tested here and do not touch this branch: one preserving union default role members, one on the union role variable in v2, and one on role-union delete with root. If there is an existing report we missed, please close this as a duplicate.

Found by TrueCourse running the product's published documentation against a live instance; the full transcript (requests, both raw response texts compared byte for byte, and the `roleMode` readings before and after the flip) is available on request.

yzxcj797:
## What?

Under the default system role mode, `X-Role: __union__` is silently rewritten to the user's first role: the request answers 200 under a role the caller never named, `roles:check` reports the substituted role, and the incoming header is rewritten in place — so neither the caller nor downstream server code can learn that a substitution happened (#10399).

Closes #10399

## Why the fix is a signal, not a refusal

The issue lists refusal (401/400) as the first preference, and the compatibility alternative — "the response should carry a signal that the substitution happened" — as acceptable if the degradation is deliberate. It is deliberate: `#1907` ("should currentRole not be `__union__` when default role mode") pins the first-role resolution for a login token minted with the union role, and the token's `roleName` is injected into `ctx.headers['x-role']` by `auth.check()` *before* this middleware runs — so an explicit caller header and the login-token path are indistinguishable here. Refusing one would refuse the other and break the documented login flow.

## How?

Keep the resolution exactly as-is and stop the silence:

- `ctx.set('X-Role-Substituted', '__union__')` — the caller can finally see that the answer came from a different role than the one named;
- `ctx.state.requestedRole = '__union__'` — downstream server code can see what was actually requested.

No status code, role resolution, or header-rewrite behavior changes.

## Testing

Two new cases in `union-role.test.ts` (16/16 pass, including the untouched `#1907`):

- an explicit `X-Role: __union__` request under the default mode answers 200 with the first role **and** the `X-Role-Substituted` signal; a request naming a held role carries no signal;
- under `allow-use-union`, the union request is answered as the union, with no signal.

**Differential**: on `main` the signal assertion fails with `expected undefined to be '__union__'` — the exact silent substitution from the report. (Note: `setCurrentRole.test.ts` cannot run in my local sandbox due to a Windows vite/builtin-module resolution quirk — verified to fail identically on pristine `main`, so it is environmental; the middleware's full behavior is covered by the union-role suite above.)

### DIFF
--- a/packages/core/actions/src/__tests__/destroyed-collection-404.test.ts
+++ b/packages/core/actions/src/__tests__/destroyed-collection-404.test.ts
@@ -1,0 +1,63 @@
+/**
+ * This file is part of the NocoBase (R) project.
+ * Copyright (c) 2020-2024 NocoBase Co., Ltd.
+ * Authors: NocoBase Team.
+ *
+ * This project is dual-licensed under AGPL-3.0 and NocoBase Commercial License.
+ * For more information, please refer to: https://www.nocobase.com/agreement.
+ */
+
+import { registerActions } from '@nocobase/actions';
+import { mockServer, MockServer } from './index';
+
+describe('destroyed collection', () => {
+  let app: MockServer;
+
+  beforeEach(async () => {
+    app = mockServer();
+    registerActions(app);
+
+    app.collection({
+      name: 'tcdestroy',
+      fields: [{ type: 'string', name: 'name' }],
+    });
+
+    await app.db.sync();
+  });
+
+  afterEach(async () => {
+    await app.destroy();
+  });
+
+  it('answers 404 for a list request once the collection is destroyed', async () => {
+    const before = await app.agent().resource('tcdestroy').list();
+    expect(before.status).toBe(200);
+
+    app.db.removeCollection('tcdestroy');
+
+    const response = await app.agent().resource('tcdestroy').list();
+    expect(response.status).toBe(404);
+  });
+
+  it('answers 404 for a get request once the collection is destroyed', async () => {
+    const before = await app.agent().resource('tcdestroy').list();
+    expect(before.status).toBe(200);
+
+    app.db.removeCollection('tcdestroy');
+
+    const response = await app.agent().resource('tcdestroy').get({ filterByTk: 1 });
+    expect(response.status).toBe(404);
+  });
+
+  it('answers 404 for a create request once the collection is destroyed', async () => {
+    app.db.removeCollection('tcdestroy');
+
+    const response = await app.agent().resource('tcdestroy').create({ values: { name: 'x' } });
+    expect(response.status).toBe(404);
+  });
+
+  it('still answers 404 for a collection that never existed', async () => {
+    const response = await app.agent().resource('tcneverexisted').list();
+    expect(response.status).toBe(404);
+  });
+});
--- a/packages/core/actions/src/utils.ts
+++ b/packages/core/actions/src/utils.ts
@@ -26,16 +26,28 @@ export function pageArgsToLimitArgs(
 export function getRepositoryFromParams(ctx: Context) {
   const { resourceName, sourceId, actionName } = ctx.action;
 
+  let repository: Repository | MultipleRelationRepository | undefined;
+
   if (sourceId === '_' && ['get', 'list'].includes(actionName)) {
     const collection = ctx.db.getCollection(resourceName);
-    return ctx.db.getRepository<Repository>(collection.name);
+    repository = collection ? ctx.db.getRepository<Repository>(collection.name) : undefined;
+  } else if (sourceId) {
+    repository = ctx.db.getRepository<MultipleRelationRepository>(resourceName, sourceId);
+  } else {
+    repository = ctx.db.getRepository<Repository>(resourceName);
   }
 
-  if (sourceId) {
-    return ctx.db.getRepository<MultipleRelationRepository>(resourceName, sourceId);
+  // A collection that was destroyed moments ago leaves its resourcer route
+  // behind: the request still reaches the action, and getRepository returns
+  // undefined. Answer 404 here — the same answer this API gives for a
+  // collection that never existed — instead of letting the action crash
+  // with a TypeError ("Cannot read properties of undefined (reading
+  // 'collection')") and surface as a 500 (#10397).
+  if (!repository) {
+    ctx.throw(404, `collection ${resourceName} does not exist`);
   }
 
-  return ctx.db.getRepository<Repository>(resourceName);
+  return repository;
 }
 
 export function RelationRepositoryActionBuilder(method: 'remove' | 'set') {
--- a/packages/plugins/@nocobase/plugin-acl/src/server/__tests__/union-role.test.ts
+++ b/packages/plugins/@nocobase/plugin-acl/src/server/__tests__/union-role.test.ts
@@ -518,17 +518,52 @@ describe('union role: full permissions', async () => {
     expect(createRoleResponse.statusCode).toBe(200);
   });
 
-  it('should currentRole not be __union__ when default role mode #1907', async () => {
+  it('should currentRole not be __union__ when default role mode #1907', async () => {
+    const rootAgent = await app.agent().login(rootUser);
+    await rootAgent.resource('roles').setSystemRoleMode({
+      values: {
+        roleMode: SystemRoleMode.default,
+      },
+    });
+    agent = await app.agent().login(user, UNION_ROLE_KEY);
+    const createRoleResponse = await agent.resource('roles').check();
+    expect(createRoleResponse.statusCode).toBe(200);
+    expect(createRoleResponse.body.data.role).not.toBe(UNION_ROLE_KEY);
+  });
+
+  it('should signal the first-role substitution for __union__ under the default role mode #10399', async () => {
     const rootAgent = await app.agent().login(rootUser);
     await rootAgent.resource('roles').setSystemRoleMode({
       values: {
         roleMode: SystemRoleMode.default,
       },
     });
-    agent = await app.agent().login(user, UNION_ROLE_KEY);
-    const createRoleResponse = await agent.resource('roles').check();
-    expect(createRoleResponse.statusCode).toBe(200);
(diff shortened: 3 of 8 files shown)


truecourse-agent:
I reported #10399, where `X-Role: __union__` under the default role mode was quietly rewritten to the user's first role and answered 200 with nothing to say a substitution had happened.

My check asserts a refusal, so on this branch it still fails. That is my check being narrower than the issue rather than a problem with the PR. The issue named a refusal as the first preference and a signal as an acceptable alternative, and this takes the second.

✅ I re-ran it asserting the response header instead, and `X-Role-Substituted` is there. The substitution is no longer silent, which is what was asked for. I will widen my own check.

One thing you may not have intended. This branch also carries the field-interface default-type fix for #10398, the same change as #10400, which accounts for most of its size. If that was not deliberate it may be worth rebasing so this PR is only the role change.

PR #10411, opened 2026-08-25
fix(workflow): stop static schedule from running past endsOn

Tespera:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Motivation

A schedule workflow in static mode keeps running **once past its `endsOn`**.

`StaticScheduleTrigger.getNextTime()` guards `endsOn` only against the *current* time:

```ts
const endTime = config.endsOn ? parseDateWithoutMs(config.endsOn) : null;
if (endTime && endTime < timestamp) {   // only "has endsOn already passed?"
  return null;
}
// ... the occurrence computed below may itself be beyond endsOn, and is scheduled anyway
```

So whenever `endsOn` is still ahead but the next occurrence falls beyond it, that occurrence is scheduled and executed. It self-terminates on the following pass (by then "now" is past `endsOn`), so the effect is exactly one extra run — but how late that run happens is bounded only by the repeat period:

| repeat | endsOn | last run actually executed | overrun |
| --- | --- | --- | --- |
| `0 30 10 * * 5` (weekly, Fri) | Wed | Fri | +2 days |
| `0 30 10 * * 2,5` (Tue/Fri) | Sat | next Tue | +3 days |
| `0 0 10 1 * *` (monthly, 1st) | Sep 15th | **Oct 1st** | **+16 days** |
| `86400000` (every 24h) | Aug 26th 12:00 | Aug 27th 09:00 | +21 hours |

Both repeat kinds are affected, cron and numeric interval alike.

The sibling implementation already gets this right. `DateFieldScheduleTrigger.getRecordNextTime()` re-checks the computed time against `endsOn` in both branches:

```ts
if (typeof repeat === 'number') {
  nextTime = timestamp + repeat - ((timestamp - startTime) % repeat);
  ...
  if (endTime && endTime < nextTime) {
    logger.debug(`[Schedule on date field] getNextTime: nextTime is after endsOn`);
    return null;
  }
} else if (typeof repeat === 'string') {
  nextTime = getCronNextTime(repeat, now);
  ...
  if (endTime && endTime < nextTime) { ... return null; }
}
```

This PR brings the static trigger in line with it, so the two schedule modes behave the same way.

### Description

After computing the next occurrence, both branches now return `null` when it lies beyond `endsOn`. The comparison is `endTime < next`, matching `DateFieldScheduleTrigger` — `endsOn` stays **inclusive**, so an occurrence landing exactly on it still runs.

Risks: low, and symmetrical for both repeat kinds. The only behaviour that disappears is a run that the user explicitly asked not to happen. Nothing changes when `endsOn` is unset, and nothing changes for occurrences at or before `endsOn`.

Testing suggestions: `mode-static.test.ts` had **no** `endsOn` coverage at all (`mode-date-field.test.ts` exercises it in eight places), which is why this went unnoticed. Three cases are added:

- `should not trigger after endsOn on cron repeat` — every 2s, `endsOn` at +3s; expects one run at +2s (currently two).
- `should not trigger after endsOn on interval repeat` — same with `repeat: 2000`, covering the numeric branch (currently two).
- `should trigger on an occurrence exactly at endsOn` — guards the inclusive boundary; passes before and after the fix.

Verified locally against Postgres 17: 17/17 pass with the fix; reverting only the trigger while keeping the tests makes the first two fail with `expected 2 to be 1`.

### Related issues

Same function as #10410 (extra trigger at `startsOn` for cron repeat). This branch is based on `main` and does not include that change, so a rebase may be needed depending on merge order.

### Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix schedule workflows in static mode running one more time after `endsOn` |
| 🇨🇳 Chinese | 修复定时任务（静态模式）在「结束于」之后仍会多执行一次的问题 |

### Docs

| Language   | Link |
| ---------- | --------- |
| 🇺🇸 English | |
| 🇨🇳 Chinese | |

### Checklists

### DIFF
--- a/packages/plugins/@nocobase/plugin-workflow/src/server/__tests__/triggers/schedule/mode-static.test.ts
+++ b/packages/plugins/@nocobase/plugin-workflow/src/server/__tests__/triggers/schedule/mode-static.test.ts
@@ -244,6 +244,155 @@ describe('workflow > triggers > schedule > static mode', () => {
       expect(new Date(executions[0].context.date).getTime()).toBe(start.getTime());
     });
 
+    it('should not trigger after endsOn on cron repeat', async () => {
+      await sleepToEvenSecond();
+
+      const base = new Date();
+      base.setMilliseconds(0);
+      // repeats every 2 seconds, ends 3 seconds later: only the +2s occurrence is within `endsOn`
+      const workflow = await WorkflowRepo.create({
+        values: {
+          enabled: true,
+          type: 'schedule',
+          config: {
+            mode: 0,
+            startsOn: base.toISOString(),
+            repeat: '*/2 * * * * *',
+            endsOn: new Date(base.getTime() + 3000).toISOString(),
+          },
+        },
+      });
+
+      await sleep(6000);
+
+      const executions = await workflow.getExecutions();
+      expect(executions.length).toBe(1);
+      expect(new Date(executions[0].context.date).getTime()).toBe(base.getTime() + 2000);
+    });
+
+    it('should not trigger after endsOn on interval repeat', async () => {
+      await sleepToEvenSecond();
+
+      const base = new Date();
+      base.setMilliseconds(0);
+      const workflow = await WorkflowRepo.create({
+        values: {
+          enabled: true,
+          type: 'schedule',
+          config: {
+            mode: 0,
+            startsOn: base.toISOString(),
+            repeat: 2000,
+            endsOn: new Date(base.getTime() + 3000).toISOString(),
+          },
+        },
+      });
+
+      await sleep(6000);
+
+      const executions = await workflow.getExecutions();
+      expect(executions.length).toBe(1);
+      expect(new Date(executions[0].context.date).getTime()).toBe(base.getTime() + 2000);
+    });
+
+    it('should trigger on an occurrence exactly at endsOn', async () => {
+      await sleepToEvenSecond();
+
+      const base = new Date();
+      base.setMilliseconds(0);
+      // `endsOn` is inclusive: an occurrence landing exactly on it still runs
+      const workflow = await WorkflowRepo.create({
+        values: {
+          enabled: true,
+          type: 'schedule',
+          config: {
+            mode: 0,
+            startsOn: base.toISOString(),
+            repeat: '*/2 * * * * *',
+            endsOn: new Date(base.getTime() + 2000).toISOString(),
+          },
+        },
+      });
+
+      await sleep(4000);
+
+      const executions = await workflow.getExecutions();
+      expect(executions.length).toBe(1);
+      expect(new Date(executions[0].context.date).getTime()).toBe(base.getTime() + 2000);
+    });
+
+    it('should not trigger when startsOn is after endsOn on cron repeat', async () => {
+      await sleepToEvenSecond();
+
+      const base = new Date();
+      base.setMilliseconds(0);
+      // a `startsOn` later than `endsOn` is contradictory: nothing should run, not even the start itself
+      const workflow = await WorkflowRepo.create({
+        values: {
+          enabled: true,
+          type: 'schedule',
+          config: {
+            mode: 0,
+            startsOn: new Date(base.getTime() + 3000).toISOString(),
+            repeat: '*/2 * * * * *',
+            endsOn: new Date(base.getTime() + 1000).toISOString(),
+          },
+        },
+      });
+
+      await sleep(6000);
+
+      const executions = await workflow.getExecutions();
+      expect(executions.length).toBe(0);
+    });
+
+    it('should not trigger when startsOn is after endsOn on interval repeat', async () => {
+      await sleepToEvenSecond();
+
+      const base = new Date();
+      base.setMilliseconds(0);
+      const workflow = await WorkflowRepo.create({
+        values: {
+          enabled: true,
+          type: 'schedule',
+          config: {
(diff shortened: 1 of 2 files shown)


mytharcher (changes requested):
Request changes: the main endsOn overrun is real and the added checks correctly cover next occurrences beyond a future endsOn. However, getNextTime() still returns startTime before reading or validating endsOn (the early return at lines 66-68 of the PR version). Therefore a configuration with startsOn after endsOn still schedules and executes at startsOn for both numeric and cron repeats. This is an endsOn violation and is not prevented by the new tests; please add a regression case for startsOn > endsOn and ensure the early-start path also respects the end bound (or reject/normalize that invalid configuration). Local PostgreSQL verification: the three added endsOn tests pass with the PR change (3/3).

Tespera:
Thanks, confirmed and fixed in 718e799f.

You were right that the early return at the top of `getNextTime()` bypassed `endsOn` entirely. One addition: since that return does not look at `repeat` at all, the violation is not limited to numeric and cron repeats — a **non-repeating** schedule with `startsOn` after `endsOn` also fired at `startsOn`.

`endsOn` is now resolved before the early return, and a start beyond it yields no schedule:

```ts
const startTime = parseDateWithoutMs(config.startsOn);
const endTime = config.endsOn ? parseDateWithoutMs(config.endsOn) : null;
if (startTime > timestamp) {
  return endTime && endTime < startTime ? null : startTime;
}
```

I went with "do not schedule" rather than rejecting the configuration at write time, to keep this PR limited to the trigger. Happy to move it to validation instead if you prefer that.

Three regression cases added, covering all three paths:

- `should not trigger when startsOn is after endsOn on cron repeat`
- `should not trigger when startsOn is after endsOn on interval repeat`
- `should not trigger when startsOn is after endsOn without repeat`

Verified against PostgreSQL 17 — 20/20 pass with the change. Reverting only `StaticScheduleTrigger.ts` to `main` while keeping the tests fails exactly the five new cases and no others:

```
✗ should not trigger after endsOn on cron repeat                        expected 2 to be 1
✗ should not trigger after endsOn on interval repeat                    expected 2 to be 1
✗ should not trigger when startsOn is after endsOn on cron repeat       expected 1 to be +0
✗ should not trigger when startsOn is after endsOn on interval repeat   expected 1 to be +0
✗ should not trigger when startsOn is after endsOn without repeat       expected 1 to be +0
✓ should trigger on an occurrence exactly at endsOn                     (inclusive boundary guard)
```

mytharcher (approved):
The requested change has been addressed. `endsOn` is now resolved before the early `startsOn` return, so contradictory configurations with `startsOn > endsOn` are not scheduled for cron, numeric interval, or non-repeating schedules. The three added regression cases cover all of those paths. The original next-occurrence checks remain correct and preserve the inclusive `endsOn` boundary. I also verified that the resulting logic is compatible with #10410 when the branches are combined. Local PostgreSQL verification passed: all 6 `endsOn`-related tests passed (14 unrelated tests skipped by the title filter). No blocking issues found. Approved.

mytharcher:
Please resolve the conflicts before merge.

Tespera:
Rebased onto current `main` (which now contains #10410) — conflicts resolved, `fa069ee`.

Worth flagging how the two changes compose, since it moved which branch guards one of the cases:

```ts
const startTime = parseDateWithoutMs(config.startsOn);
const endTime = config.endsOn ? parseDateWithoutMs(config.endsOn) : null;
// ...
if (startTime > timestamp && typeof config.repeat !== "string") {
  return endTime && endTime < startTime ? null : startTime;
}
```

After #10410 a cron repeat no longer takes the early return at all, so `startsOn > endsOn` **with a cron repeat** is now caught further down instead: the cron branch computes its next occurrence from `startsOn - 1s`, so that occurrence is `>= startsOn > endsOn` and the `endTime < next` check returns `null`. The early return keeps covering the numeric and non-repeating cases. All three regression tests still assert zero executions, they just reach that result through different branches now.

On verification — I have to be precise here: my local checkout was wiped between the review and this rebase, and reinstalling the monorepo dependencies is a ~15 minute / several-GB round trip, so **I did not re-run the suite locally for this rebase**. What I did do was replay the merged `getNextTime()` against every assertion in the file, including the two cases #10410 added, using the same sleep windows the tests use:

```
✓ [#10410] startsOn in future, not matching cron   sleep=6000  expected 1 @ +4s   got 1 @ +4s
✓ [#10410] startsOn in future, matching cron       sleep=3000  expected 1 @ +2s   got 1 @ +2s
✓ [overrun] cron, endsOn=+3s                       sleep=6000  expected 1 @ +2s   got 1 @ +2s
✓ [overrun] numeric repeat, endsOn=+3s             sleep=6000  expected 1 @ +2s   got 1 @ +2s
✓ [boundary] occurrence exactly at endsOn          sleep=4000  expected 1 @ +2s   got 1 @ +2s
✓ [contradictory] startsOn > endsOn, cron          sleep=6000  expected 0         got 0
✓ [contradictory] startsOn > endsOn, numeric       sleep=6000  expected 0         got 0
✓ [contradictory] startsOn > endsOn, no repeat     sleep=5000  expected 0         got 0
```

8/8 consistent with the assertions, and `prettier --check` passes. The real run is on CI. Happy to reinstall and run the suite locally if you would rather have that before merging.

PR #10437, opened 2026-08-31
fix(plugin-workflow): ensure global request-interception workflows are triggered after restart

### LINKED ISSUE #10435: [Bug] Global request-interception workflow not triggered after restart (enabledCache empty + middleware not bound to already-used datasource)
## * Describe the bug

Global `request-interception` (Pre-action Event) workflow with `sync: true` does **not intercept** the request. The `DELETE` reaches the database and fails with the raw Postgres FK error instead of the workflow's `400` with `response-message` + `end (endStatus: -1)`.

Two independent timing bugs combine to make the documented global sync flow unreliable after a restart:

**1. `enabledCache` is empty for `request-interception` after restart.**
`@nocobase/plugin-workflow` populates `enabledCache` in `PluginWorkflowServer.onAfterStart()` via `toggle(workflow, true)`. `@nocobase/plugin-workflow-request-interceptor` registers the trigger in `PluginWorkflowRequestInterceptorServer.load()` via `registerTrigger("request-interception", ...)`. If `onAfterStart` runs before the trigger is registered, workflows with `type: "request-interception"` are not loaded into the cache. Result: `workflow.enabledCache.size === 0`, `allWorkflows === []`, `syncWorkflows === []` — the middleware calls `await next()` and never triggers the workflow.

`nb api workflow workflows sync` re-hydrates the cache temporarily, but the problem returns on next restart.

**2. Middleware never bound to `resourceManager` when the datasource is already `_used`.**
`RequestInterceptionTrigger` constructor does:
```js
workflow.app.dataSourceManager.use(this.middleware, { after: "parseVariables", before: "default" })
```
`DataSourceManager.use()` only queues the middleware. It is flushed to each `dataSource.resourceManager` when the datasource is added or when `_used` becomes true. For the default `main` datasource, `_used` is already `true` when the trigger is constructed, so the middleware is never applied to `resourceManager`. The `clientes:destroy` request never enters the interceptor.

Together, a global destroy workflow that should return `400 "No se puede borrar..."` instead lets the request through to Postgres: `500 violates foreign key constraint "movimientos_clienteid_fkey"`.

## * Environment

- NocoBase version: `2.2.1` (`@nocobase/plugin-workflow@2.2.1`, `@nocobase/plugin-workflow-request-interceptor@2.2.1`)
- Database type and version: PostgreSQL 16
- OS: Linux (Docker host)
- Deployment Methods: Docker
- Docker image version: `nocobase/nocobase:latest-full` (resolved to 2.2.1 at time of report)
- NodeJS version: `v22.23.2` (inside container)

## * How To Reproduce

1. Create two collections: `clientes` and `movimientos` with `movimientos.clienteId -> clientes.id ON DELETE RESTRICT` (or any FK with restrict).
2. Create a global sync workflow:
```json
{
  "title": "Bloquear borrado de cliente con movimientos",
  "type": "request-interception",
  "enabled": true,
  "sync": true,
  "config": { "collection": "clientes", "global": true, "actions": ["destroy"] }
}
```
3. Add nodes (in order):
   - `aggregate` `count` on `movimientos` where `clienteId = {{$context.params.filterByTk}}` (key `mu7xla118i2`)
   - `condition` `basic` `gt` `{{$jobsMapByNodeKey.mu7xla118i2}} > 0`, `rejectOnFalse: false`
   - `response-message` on branch `1`: `"No se puede borrar el cliente porque tiene {{$jobsMapByNodeKey.mu7xla118i2}} movimiento(s)..."`
   - `end` `endStatus: -1` on same branch
4. Ensure the workflow is enabled. Restart the app container (`docker compose restart app` or `down/up`).
5. Create a `clientes` record with at least one `movimientos` referencing it.
6. Call `POST /api/clientes:destroy?filterByTk=<id>` (or via `nb api resource destroy --resource clientes --filter-by-tk <id>`).

**Actual:** request is not intercepted. DB error:
```
violates foreign key constraint "movimientos_clienteid_fkey" on table "movimientos"
```
With debug logging in the trigger middleware:
```
[WF-INTERCEPT-DEBUG] middleware entered { resourceName: 'clientes', actionName: 'destroy', enabledCacheSize: 0 }
[WF-INTERCEPT-DEBUG] allWorkflows []
[WF-INTERCEPT-DEBUG] syncWorkflows { count: 0 }
```

**After running `nb api workflow workflows sync --filter-by-tk <workflowId>`** the cache is re-hydrated but the middleware is still not bound (bug #2), so interception still fails until the `resourceManager` patch is applied.

## Expected behavior

The global sync workflow should intercept `clientes:destroy` before it reaches the collection, execute the node chain, and when `count > 0` return `400` with the `response-message` body. Only when `count == 0` should the destroy proceed. The `violates foreign key` error should never be exposed to the client.

## Screenshots

N/A — API-level reproduction.

## Logs

Before workaround (after restart, before `sync`):
```
[WF-INTERCEPT-DEBUG] middleware entered { resourceName: 'clientes', actionName: 'destroy', enabledCacheSize: 0 }
[WF-INTERCEPT-DEBUG] allWorkflows []
[WF-INTERCEPT-DEBUG] syncWorkflows { count: 0, globalWorkflows: [], workflows: [] }
```

After `sync` but before middleware patch — cache has the workflow but middleware was never called (request went straight to DB).

After both workarounds (verified):
```
nb api resource destroy --resource clientes --filter-by-tk <idWithMovimientos>
→ 400 {"errors":[{"message":"No se puede borrar el cliente porque tiene 2 movimiento(s). Borrá o reasigná esos movimientos primero."}]}

nb api resource destroy --resource clientes --filter-by-tk <idWithoutMovimientos>
→ {"data":1}
```

## Workaround we used to verify

We confirmed the fix with two minimal patches (reversible, mounted as volumes):

**`Plugin.js` — re-hydrate after all plugins are loaded:**
```js
async afterStart() {
  const workflowPlugin = this.app.pm.get(WorkflowPlugin);
  const workflows = await this.db.getRepository("workflows").find({
    filter: { type: "request-interception", enabled: true, current: true }
  });
  for (const w of workflows) workflowPlugin.toggle(w, true, { silent: true });
}
```

**`RequestInterceptionTrigger.js` — bind to already-used datasources:**
```js
const directlyApplied = new WeakSet();
const applyIfMissed = (dataSource) => {
  if (!dataSource?.resourceManager || !dataSource["_used"] || directlyApplied.has(dataSource)) return;
  dataSource.resourceManager.use(this.middleware, { after: "parseVariables", before: "default" });
  directlyApplied.add(dataSource);
};
workflow.app.dataSourceManager.afterAddDataSource(applyIfMissed);
for (const ds of workflow.app.dataSourceManager.dataSources.values()) applyIfMissed(ds);
```
Plus `execute()` should forward `transaction: httpContext.state.transaction` to `repository.findOne`.

Happy to open a PR with the same changes if the approach looks good.

fede-ciliberti:
Fixes #10435

Global sync `request-interception` workflows never intercepted the request after a restart — the DELETE reached Postgres and exposed the raw FK error (`violates foreign key`).

Two timing bugs:

1. **enabledCache empty after restart** — `PluginWorkflowServer.onAfterStart` hydrates `enabledCache` before `plugin-workflow-request-interceptor` registers the trigger via `load()`, so workflows with `type=request-interception` never enter the cache. `PluginWorkflowRequestInterceptorServer.afterStart()` now re-hydrates enabled `request-interception` workflows with `toggle(..., silent:true)`.

2. **Middleware not bound when datasource already _used** — `RequestInterceptionTrigger` does `dataSourceManager.use(middleware)` after `main` is already `_used=true`, so `DataSource.middleware()` never flushes it to `resourceManager`. `DataSourceManager.use()` now also binds the middleware to already-`_used` datasources.

3. Forward the request `transaction` in `RequestInterceptionTrigger.execute()` (`httpContext.state.transaction`) to keep the read inside the request transaction.

Verified: `clientes:destroy` with movimientos → `400` with `response-message`; without movimientos → success. Previously returned `500 violates foreign key`.

### DIFF
--- a/packages/core/data-source-manager/src/data-source-manager.ts
+++ b/packages/core/data-source-manager/src/data-source-manager.ts
@@ -87,6 +87,11 @@ export class DataSourceManager {
 
   use(fn: any, options?: ToposortOptions) {
     this.middlewares.push([fn, options]);
+    for (const ds of this.dataSources.values()) {
+      if ((ds as any)._used && ds.resourceManager) {
+        ds.resourceManager.use(fn, options);
+      }
+    }
   }
 
   middleware() {
--- a/packages/plugins/@nocobase/plugin-workflow-request-interceptor/src/server/Plugin.ts
+++ b/packages/plugins/@nocobase/plugin-workflow-request-interceptor/src/server/Plugin.ts
@@ -17,6 +17,16 @@ export class PluginWorkflowRequestInterceptorServer extends Plugin {
     const workflowPlugin = this.app.pm.get(PluginWorkflowServer) as PluginWorkflowServer;
     workflowPlugin.registerTrigger('request-interception', RequestInterceptionTrigger);
   }
+
+  async afterStart() {
+    const workflowPlugin = this.app.pm.get(PluginWorkflowServer) as PluginWorkflowServer;
+    const workflows = await this.db.getRepository('workflows').find({
+      filter: { type: 'request-interception', enabled: true, current: true },
+    });
+    for (const workflow of workflows) {
+      workflowPlugin.toggle(workflow, true, { silent: true });
+    }
+  }
 }
 
 export default PluginWorkflowRequestInterceptorServer;
--- a/packages/plugins/@nocobase/plugin-workflow-request-interceptor/src/server/RequestInterceptionTrigger.ts
+++ b/packages/plugins/@nocobase/plugin-workflow-request-interceptor/src/server/RequestInterceptionTrigger.ts
@@ -274,6 +274,7 @@ export default class RequestInterceptionTrigger extends Trigger {
     if (loadNeeded) {
       target = await repository.findOne({
         filterByTk,
+        transaction: (options as any)?.httpContext?.state?.transaction ?? (options as any)?.transaction,
       });
     }
     const params = values.action === INTERCEPTABLE_ACTIONS.DESTROY ? { filterByTk } : { values: target };


mytharcher:
Thanks a lot for digging into this — the underlying problem is real, and it's a genuinely tricky one to track down. I spent some time reproducing it locally, and I ended up with a slightly different picture of the root cause, so I wanted to share what I found before we move forward. Apologies in advance for the long comment.

## What I was able to reproduce

I checked out this branch and wrote some throwaway tests against `plugin-workflow-request-interceptor`:

- **Process restart (`app.stop()` + `app.start()`) — I couldn't reproduce the failure.** On unpatched `main`, after a restart `enabledCache` still contains `{ type: 'request-interception' }`, the middleware fires, and interception works as expected. Looking at the ordering: `pm.load()` runs at `application.ts:719` inside `app.load()`, while `afterStart` is emitted later at `application.ts:925` inside `app.start()` — so triggers do appear to be registered before hydration. If you have a reproduction where this ordering is inverted, I'd love to see it; it's very possible I'm missing a code path.

- **`app.reload()` — this one reproduces reliably.** Before the reload the cache holds `['collection', 'request-interception']`; afterwards it's `[]`, and the request is no longer intercepted. The cause seems to be that `app.load({ reload: true })` calls `this.init()` (`application.ts:673`), which constructs a **brand-new** `PluginWorkflowServer` instance (I confirmed `app.pm.get('workflow') !== previousInstance`), and `afterStart` is never re-emitted — so the fresh, empty `enabledCache` never gets hydrated.

Worth noting: this affects **every** trigger type, not just `request-interception`. In my repro a plain `collection`-trigger workflow was dropped by the same reload.

## A few things I'd like to check with you

Sorry to pile on here — these are all things I'd want to understand better rather than firm objections.

**1. `toggle()` is currently private, so the branch doesn't compile.**
`workflowPlugin.toggle(workflow, true, { silent: true })` (`plugin-workflow-request-interceptor/src/server/Plugin.ts:27`) calls a `private` method (`plugin-workflow/src/server/Plugin.ts:478`, and `private toggle;` in the published `dist/server/Plugin.d.ts`). Running `tsc` gives:

```
error TS2341: Property 'toggle' is private and only accessible within class 'PluginWorkflowServer'
```

Making this work would mean widening `toggle`'s visibility in `plugin-workflow`, which feels like a public-API decision worth making deliberately.

**2. I don't think `afterStart()` is ever invoked on plugins.**
The `Plugin` base class (`packages/core/server/src/plugin.ts:44-141`) doesn't declare an `afterStart` lifecycle hook, and `PluginManager` only calls `afterAdd()` / `beforeLoad()` / `load()` / install / upgrade / enable / disable / remove (`plugin-manager.ts:367,448,469,594,610`). Grepping `packages/core` and `packages/plugins` for `plugin.afterStart()` turned up no call sites. If that's right, this method would never run even once the TS error is resolved. Registering `this.app.on('afterStart', ...)` from `load()` would be the equivalent hook — though see the next point.

**3. Would it make sense to fix this in `plugin-workflow` instead?**
Since the reload gap affects all trigger types, re-hydrating only `type: 'request-interception'` from a downstream plugin would leave `collection`, `schedule`, `action`, approval, etc. broken after every reload. Hydrating the cache from `load()` (or `afterLoad`) in `plugin-workflow` rather than only from the one-shot `afterStart` would cover everyone at once — and would let this plugin stay unchanged.

**4. The filter differs slightly from the core hydration path.**
`PluginWorkflowServer.onAfterStart` (`plugin-workflow/src/server/Plugin.ts:176-196`) toggles on `workflow.enabled` alone with no `current` predicate, and also backfills `stats` / `versionStats` for compatibility. With `current: true` added here, an enabled non-current version would silently not be re-hydrated, and the re-hydrated models wouldn't carry the `versionStats` append — so the resulting state would differ subtly from a real startup.

**5. The `data-source-manager` change may be guarding a state that can't occur.**
The `if ((ds as any)._used && ds.resourceManager)` branch (`data-source-manager.ts:90`) can only be reached if `use()` is called after a datasource has already served a request. As far as I can tell, `_used` is set solely in `DataSource.middleware()` (`data-source.ts:88-93`), which runs per-request, while every `use()` call site in the repo (`application.ts:1332-1341`, plus the trigger/plugin constructors in `RequestInterceptionTrigger.ts:224`, `ApprovalTrigger.ts:485`, `ActionTrigger.ts:63`, `plugin-verification/Plugin.ts:43`) runs during construction or `pm.load()`. Empirically I logged `mainDataSource['_used'] === undefined` immediately after `app.reload()`, since `init()` rebuilds both the `DataSourceManager` and the main `DataSource`.

Two follow-on worries if we do keep it: `use()` still pushes `fn` into `this.middlewares` *and* now eagerly registers it, so any future path that re-runs `DataSource.middleware(self.middlewares)` on the same `resourceManager` would install it twice — meaning two executions and a duplicate interception error per request. And `(ds as any)._used` reaches into an untyped flag, which our `AGENTS.md` asks us to avoid; if the behaviour is wanted, promoting `_used` to a declared field with an accessor would be safer. Given this is a core, cross-cutting module, would you be open to dropping this change unless there's a failing case behind it that I've missed?

**6. The `transaction` forwarding looks like it may be a no-op today.**
`httpContext.state.transaction` (`RequestInterceptionTrigger.ts:277`) doesn't appear to be assigned anywhere — grepping `state.transaction` across `packages/core` and `packages/plugins` (excluding `node_modules`/`dist`) returned nothing, and the interception middleware passes only `{ httpContext: context }` (`RequestInterceptionTrigger.ts:163`). So both sides of the `??` would be `undefined` and `repository.findOne` behaves exactly as before.

Also, if a transaction were ever supplied, `execute()` is reachable from the public `PluginWorkflowServer.execute()` (`plugin-workflow/src/server/Plugin.ts:610`) and from `ScheduleTrigger` (`triggers/ScheduleTrigger/index.ts:70`), where execution can be deferred — a request-scoped transaction would then throw `Transaction cannot be used after commit` once the HTTP request has finished. If we do want transaction forwarding, gating it to the synchronous pre-action middleware path would be safer. Minor: the two `as any` casts aren't needed — `EventOptions` (`Dispatcher.ts:53-68`) already has `[key: string]: any` and intersects `Transactionable`.

**7. Could we add a regression test?**
Since this touches a shared core module, a test would help a lot. The plugin already has a good harness (`src/server/__tests__/trigger.test.ts`, `getApp` from `@nocobase/plugin-workflow-test`). I wrote one while investigating that asserts interception still fires after `app.reload()` — it fails on `main`, so it captures exactly the case this PR is about. Happy to share it if that's useful.

## Suggested direction

If my reading is correct, the defect worth fixing is: **`app.reload()` recreates `PluginWorkflowServer` with an empty `enabledCache` and never re-hydrates it, disabling all workflow triggers until the next real process start.** That would put the fix in `packages/plugins/@nocobase/plugin-workflow/src/server/Plugin.ts` — hydrating from `load()` / `afterLoad` instead of only from `afterStart` — with a reload regression test, and would let the `data-source-manager` and transaction changes be dropped.

I could easily be wrong on any of the above, especially if you're seeing the failure in a setup I didn't reproduce — please push back. And thanks again for spotting this; the reload gap is a real bug and it's good that it's been surfaced. Happy to pair on the `plugin-workflow`-side version if you'd like.

PR #10467, opened 2026-09-06
fix: resolve popup variables in form linkage rules

gchust:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Motivation

Non-admin users could not resolve configured variables when form linkage rules or referenced events were stored on a different model from the runtime context.

### Description 

Include direct and referenced form-grid linkage rules and persisted RunJS dependencies in the variable allow-list. Preserve the configuration owner for forwarded reference events, and include the validated target model's own configuration and form-grid rules without changing its record context. No permission changes or data migration.

The latest follow-up changes one production file (+11/-3). It replaces the synthetic form-request test with authenticated HTTP coverage, adds delayed-response form-rendering coverage, and fixes the CI permission test's click-before-loading-completes race with a three-line wait. Tests drain automatic linkage refreshes between cases rather than changing default-value behavior: edit-mode defaults do not overwrite existing records; override values still apply.

Verification of the exact committed source tree: [regression run](https://github.com/nocobase/nocobase/actions/runs/34138712346).

| Suite | Files | Passed | Skipped |
| --- | ---: | ---: | ---: |
| Variable server tests | 19 | 364 | 0 |
| Core Flow Engine | 124 | 1634 | 1 existing |
| Template client | 11 | 97 | 0 |
| Form runtime and popup linkage | 14 | 155 | 0 |
| UI layout client | 9 | 186 | 0 |

ESLint and `git diff --check` passed. The permission test file also passed three consecutive reruns (8 tests each). These are the affected regression suites, not a claim that the entire repository or every database/platform matrix has been rerun. The live reported instance was not accessed.

### Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fixed non-admin variable resolution in form linkage rules, referenced templates, and scripts. |
| 🇨🇳 Chinese | 修复非管理员用户在表单联动规则、引用模板及脚本中的变量解析问题。 |

### Docs

| Language   | Link |
| ---------- | --------- |
| 🇺🇸 English |      |
| 🇨🇳 Chinese |    |

### Checklists
- [ ] If documentation was changed, the corresponding files in all other languages have been updated to keep them in sync (or this change does not affect docs)

### DIFF
--- a/packages/core/client-v2/src/flow/models/blocks/form/__tests__/popupLinkage.test.tsx
+++ b/packages/core/client-v2/src/flow/models/blocks/form/__tests__/popupLinkage.test.tsx
@@ -1,0 +1,175 @@
+/**
+ * This file is part of the NocoBase (R) project.
+ * Copyright (c) 2020-2024 NocoBase Co., Ltd.
+ * Authors: NocoBase Team.
+ *
+ * This project is dual-licensed under AGPL-3.0 and NocoBase Commercial License.
+ * For more information, please refer to: https://www.nocobase.com/agreement.
+ */
+
+import React from 'react';
+import { act, render, screen, waitFor, cleanup } from '@testing-library/react';
+import { Form, Input } from 'antd';
+import { afterEach, expect, it, vi } from 'vitest';
+import { FlowEngine, SingleRecordResource } from '@nocobase/flow-engine';
+import { generateFlowModelRdFromToken } from '@nocobase/utils/client';
+import { CreateFormModel, EditFormModel, FormGridModel, FormItemModel, FormComponent } from '../../../..';
+import { fieldLinkageRules, linkageAssignField } from '../../../../actions/linkageRules';
+
+afterEach(cleanup);
+
+it.each([
+  { use: 'CreateFormModel', mode: 'default', expected: 'STAFF-001' },
+  { use: 'EditFormModel', mode: 'default', expected: '' },
+  { use: 'EditFormModel', mode: 'override', expected: 'STAFF-001' },
+])('$use respects $mode linkage after a delayed popup variable response', async ({ use, mode, expected }) => {
+  const uid = `popup-${use}-${mode}`;
+  const engine = new FlowEngine();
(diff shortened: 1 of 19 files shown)


gchust:
The CI failure came from an incomplete `FlowModelRepository` test double in `variables.resolve.external-data-source.test.ts`. The new linkage allow-list path legitimately calls `findModelNodeSnapshotByParentId()` for form nodes, and the real repository implements that method; this unrelated unit test only mocked `findModelNodeSnapshotById()`.

Fixed in `b05578ebec32a3493ef13149859c1ac890c3705a` by adding:

```ts
findModelNodeSnapshotByParentId: async () => null,
```

to that fixture. Returning `null` accurately represents that its synthetic form has no grid child, while keeping the production repository contract strict.

gchust:
Found CI failure cause. The new form linkage allow-list path calls getFlowModelChildNode(), which uses FlowModelRepository.findModelNodeSnapshotByParentId(). The external-data-source tests use a mocked FlowModelRepository without this method, so the regression test fails before reaching the intended assertion.

Minimal fix: make getFlowModelChildNode() compatible with repositories that only provide findModelNodeSnapshotById(), or update the test mock to provide findModelNodeSnapshotByParentId(). Prefer the helper fallback because the production code should tolerate existing lightweight repository implementations used by tests:

```ts
const repository = getFlowModelRepository(ctx) as FlowModelRepository & {
  findModelNodeSnapshotByParentId?: (parentId: string, options?: { subKey?: string }) => Promise<FlowModelNodeSnapshot | null>;
};

const load = repository.findModelNodeSnapshotByParentId
  ? repository.findModelNodeSnapshotByParentId(parentUid, { subKey })
  : repository.findModelNodeSnapshotById(parentUid).then((parent) =>
      parent?.subModels?.[subKey] ?? null,
    );
```

Alternatively, add the missing mock method in variables.resolve.external-data-source.test.ts. The production change itself is correct; this is only a compatibility issue introduced by the new child lookup.

PR #10524, opened 2026-09-18
fix(plugin-workflow): validate aggregate target configuration

### LINKED ISSUE #10514: Aggregate node: selecting an associated collection does not populate target collection or show “Field to aggregate”
## * Describe the bug

When configuring an Aggregate node to sum a field from related records, selecting the relation appears to work, but Field to aggregate does not appear.

The saved node configuration contains the association but is missing collection and params.field. Execution then fails.

## * Environment
NocoBase version: 2.4.0-alpha5
Workflow: custom action event, triggered by a table button
Aggregate target: data of associated collection
Relation: many-to-many

- Database type and version: 17.11
- OS: fedora
- Deployment Methods:Create-nocobase-app
- NodeJS version: v24.18.0

## * How To Reproduce

1. Create a collection `re_type` with a numeric field `score`.
2. Create a collection `rel_log` with:
   - A many-to-many relation `rel1` targeting `re_type`.
   - A numeric field `log_score`.
3. Create a custom action workflow for `rel_log`, triggered by a table button.
4. Add an Aggregate node:
   - **Aggregator function:** Sum
   - **Target type:** Data of associated collection
   - **Association:** Trigger variables → Trigger data → rel1
5. Observe that **Field to aggregate** does not appear.
6. Save, enable, and trigger the workflow.

Clearing and reselecting `rel1` in a new workflow version does not resolve the issue.

### Actual behavior

The saved Aggregate configuration is:

```json
{
  "aggregator": "sum",
  "associated": true,
  "precision": 2,
  "association": {
    "name": "rel1",
    "associatedKey": "{{$context.data.id}}",
    "associatedCollection": "rel_log"
  }
}
```

Execution fails with:

```text
Cannot destructure property 'collectionManager' of 'this.workflow.app.dataSourceManager.dataSources.get(...)' as it is undefined.
```

### Expected behavior

Selecting `rel1` should automatically resolve its target collection, `re_type`, and display **Field to aggregate**, allowing selection of `score`.

The editor should also prevent saving an incomplete aggregate configuration.

### Confirmed workaround

Setting the following complete configuration through the API makes the workflow work:

```json
{
  "collection": "re_type",
  "aggregator": "sum",
  "associated": true,
  "precision": 2,
  "association": {
    "name": "rel1",
    "associatedKey": "{{$context.data.id}}",
    "associatedCollection": "rel_log"
  },
  "params": {
    "field": "score"
  }
}
```

The working and failing nodes look similar in the editor, but their saved configurations differ by the missing target collection and aggregate field.

lorenzozanee:
### This is a ...
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Others

### Motivation
An aggregate workflow node could accept an incomplete saved configuration when an associated collection was selected. The missing target collection or aggregate parameters later caused the server execution path to use an undefined data-source manager.

### Description
Require the aggregate instruction configuration to include both the target collection and aggregate parameters, while retaining the existing required aggregate field and optional filter behavior. Add regression coverage for configurations missing each required shape.

Risk is limited to server-side validation of malformed aggregate configurations. Valid ordinary and associated aggregate configurations keep their existing behavior.

Focused server tests and ESLint were attempted but could not run because the required dependencies and commands are not installed in the worktree. `git diff --check` passed. Upstream CI remains required.

### Related issues
Fixes #10514

### Showcase
Not applicable; this is a server-side validation fix.

### Changelog

| Language   | Changelog |
| ---------- | --------- |
| English | Reject incomplete aggregate workflow configurations. |
| Chinese | Reject incomplete aggregate workflow configurations. |

### Docs

| Language   | Link |
| ---------- | --------- |
| English | Not needed for this bug fix. |
| Chinese | Not needed for this bug fix. |

### Checklists
- [ ] All changes have been self-tested and work as expected


mytharcher:
Only add validation rule (required) for the configuration items will lead to failure of creating aggregate node on UI.

The `collection` will not be alway set in usage, especially for association mode. Neither `params`, because no parameters is set when creating.

So, `required` rule should not be add to both of them.

